PR #31, opened 2017-02-15
feat(widget-message-meet): Add event emitting and browser globals

bzang:
Contains initial support for callbacks and eventing. May need adjustments for better payload structure.


packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -252,6 +250,7 @@ export class MessageWidget extends Component {
    */
   listenToNewActivity(conversationId, spark) {
     const {props} = this;
+    const handleEvent = this.handleEvent;
```
bzang: good catch

PR #37, opened 2017-02-20
feat(widget-message-meet): navigation via the activity menu

adamweeks:


packages/node_modules/@ciscospark/react-component-avatar/src/styles.css
```diff
@@ -1,23 +1,51 @@
...
 }
 
 .avatar-self {
-  font-size: 34px;
+  font-size: 24px;
   color: #04c9ef;
 }
+
+.small .avatar-letter {
+  font-size: 10px;
```
adamweeks: The `font-size` for the avatar letter was the reason I had to go with the class name sizes. It works fine with just the avatar image to set it to fill the space, but not on the fonts.

packages/node_modules/@ciscospark/widget-message-meet/src/container.js
```diff
@@ -69,41 +101,68 @@ export class MessageMeetWidget extends Component {
...
+    // Don't remove message component from DOM, just hide its display
+    const messageComponentClassNames = classNames(`message-component-wrapper`, styles.messageComponentWrapper, showMessage ? `` : styles.hiddenComponent);
 
-    return (
-      <div className={classNames(`widget-message-meet`, styles.widgetMessageMeet)}>
-        <div className={classNames(`banner`, styles.banner)} />
-        <div className={classNames(`widget-message-meet-inner`, styles.widgetMessageMeetInner)}>
-          <div className={classNames(`title-bar-wrapper`, styles.titleBarWrapper)}>
+    // Don't load widgets until we have spark loaded and info on the toUser
+    if (spark && toUser && toUser.id) {
```
adamweeks: If we don't have a `toUser` and we try to load the meet widget, we get errors.

packages/node_modules/@ciscospark/react-component-avatar/src/index.js
```diff
@@ -18,7 +18,7 @@ function Avatar({name, image, isSelfAvatar}) {
   }
 
   return (
-    <div className={classNames(`avatar`, styles.avatar)} style={backgroundStyle}>
+    <div className={classNames(`avatar`, styles.avatar, size, styles[size])} style={backgroundStyle}>
```
bzang: Instead of using just the `size` string, can we prefix with something like `avatar-`? Just having a `medium` class makes it really easy for styles to collide.

packages/node_modules/@ciscospark/react-component-icon/src/index.js
```diff
@@ -14,6 +14,7 @@ export const ICON_TYPE_MESSAGE = `icon-cisco-message`;
 export const ICON_TYPE_RIGHT_ARROW = `icon-cisco-right-arrow`;
 export const ICON_TYPE_VIDEO_OUTLINE = `icon-cisco-video-outline`;
 export const ICON_TYPE_EXIT = `icon-cisco-exit`;
+export const ICON_TYPE_WAFFLE = `icon-cisco-waffle-pad`;
```
bzang: lol, it's actually called `waffle-pad` ?!?!

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
```diff
@@ -4,20 +4,31 @@ import classNames from 'classnames';
...
   );
 }
 
 TitleBar.propTypes = {
+  hasMenu: PropTypes.bool,
   image: PropTypes.string,
-  name: PropTypes.string.isRequired
+  name: PropTypes.string.isRequired,
+  onMenuClick: PropTypes.func
 };
```
bzang: I'm still of the opinion that there should be 2 components here. It just seems like this is trying to do too much, but I think that's just me being ornery.

packages/node_modules/@ciscospark/widget-message-meet/src/container.js
bzang: Yea, i'm wondering if we should just pass the `toUser` object to the child widgets and that'll give them all the info they need.

adamweeks: I think a `toUser` prop would be good, but we would still need a fallback for email/id if someone wanted to use the child widgets without fetching the user data before loading it.

packages/node_modules/@ciscospark/react-component-icon/src/index.js
adamweeks: Yup.

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
```diff
@@ -0,0 +1,72 @@
...
+import {ICON_TYPE_EXIT, ICON_TYPE_MESSAGE, ICON_TYPE_VIDEO_OUTLINE} from '@ciscospark/react-component-icon';
+
+import styles from './styles.css';
+
+export default function ActivityMenu(props) {
+  const {
+    onExit,
+    onMeet,
+    onMessage
+  } = props;
```
adamweeks: Yea, I was going to mention adding that when I got logged out, lol.

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  function handleMessage(e) {
+    e.preventDefault();
+    onMessage();
+  }
+
+  return (
+    <div className={classNames(`activity-menu`, styles.activityMenu)}>
+      <div className={classNames(`exit`, styles.exit)}>
```
adamweeks: Sure

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
```diff
@@ -0,0 +1,72 @@
...
+          <div className={classNames(`button-container`, styles.buttonContainer)}>
+            <div className={classNames(`button-item`, styles.buttonItem)}>
+              <Button iconType={ICON_TYPE_VIDEO_OUTLINE} onClick={handleMeet} title={`Call`} />
+            </div>
+          </div>
+          <div className={classNames(`menu-item-label`, styles.menuItemLabel)}>
+            {`Call`}
+          </div>
+        </div>
+      </div>
```
adamweeks: I also didn't i18n these. Should I fix that now or wait for merge?

packages/node_modules/@ciscospark/react-component-activity-menu/src/styles.css
```diff
@@ -0,0 +1,92 @@
+.activity-menu {
+  width: 280px;
```
adamweeks: Good call, will look into it.

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
adamweeks: What should the components be? It is literally a button with a click handler that either shows or it doesn't. How would you split it up?

Like: `<TitleBarWithMenu>` and `<TitleBar>`?

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
bzang: I think it makes sense to just leave it as is for now.

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
bzang: Or have a button that's attached a level up. I think I'm leaning this way because the way I think of components is isolating concerns as much as possible and this case feels like it should be 2 different components due to the fact that the menu button and the title bar are 2 separate features. Based on the zeplin comps for calling there is also a situation where the title bar isn't present but the menu button is available.

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.test.js
```diff
@@ -0,0 +1,21 @@
...
+    <ActivityMenu
+      onExit={onExit}
+      onMeet={onMeet}
+      onMessage={onMessage}
+    />
+  );
+
+  it(`renders properly`, () => {
+    expect(component).toMatchSnapshot();
+  });
```
adamweeks: We're not doing any click testing on any of the other components presently. Did you want me to implement `Enzyme` for this component?

bzang: Ahhh right. I was thinking we can test side effects, but that doesn't quite work here. Given that this PR is pretty darn important. Maybe we start adding Enzyme based tests in a future feature and then dedicate some time going back through the components to do some cleanup and add tests.

adamweeks: Yea, I think we would benefit from having the enzyme tests in anything that had click handlers and events.

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
adamweeks: Yea, that's true. I'll make the menu button a floating item that lives above the title bar.

adamweeks:
Ok @bzang, ready for round two

tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "compilerOptions": {
+    "experimentalDecorators": true,
```
adamweeks: This is to not get typescript warnings on the `@autobind` decorators. (VS Code uses typescript to generate JS warnings)

PR #40, opened 2017-02-21
feat(widget-meet): Call controls

bzang:
Added call controls, and some additional components to support the controls.


packages/node_modules/@ciscospark/widget-meet/src/components/call-active/index.js
```diff
@@ -0,0 +1,47 @@
...
+import Video from '@ciscospark/react-component-video';
+import styles from './styles.css';
+
+export default function ActiveCall({
+  intl,
+  onHangupClick,
+  localMediaStreamUrl,
+  remoteMediaStreamUrl
+}) {
+  const connectedClass = remoteMediaStreamUrl ? styles.callConnected : ``;
```
ianwremmel: would `const connectedClass = remoteMediaStreamUrl && styles.callConnected;` work?

packages/node_modules/@ciscospark/widget-meet/src/messages.js
```diff
@@ -0,0 +1,29 @@
+/*
+ * WidgetMeet Messages
+ *
+ * This contains all the text for the FeaturePage component.
+ */
+import {defineMessages} from 'react-intl';
+
+export default defineMessages({
```
ianwremmel: doesn't defining the defaults this far from the element defeat the purpose of react-intl? I thought the goal was more to define defaults at site of use, that way anyone looking at the code for the first time would understand the intent.

packages/node_modules/@ciscospark/widget-meet/src/components/call-active/index.js
bzang: Yup, had a brain fart there.

packages/node_modules/@ciscospark/widget-meet/src/messages.js
bzang: I'm still conflicted here. I inherently don't like having to include `react-intl` into what should be small self contained react components... it just ads a huge amount of overhead. I'm not sure how to resolve that.

The approach here is that we have a generic `button-control` component that just takes a `button` array as props to build up the presentational elements. That's why the i18n text is here in the container.

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
```diff
@@ -28,38 +29,33 @@ export default function ActivityMenu(props) {
     onMessage();
   }
 
+  const buttons = [
```
adamweeks: I was thinking we could even pass this array as a prop to the menu, that way the top level would control which menu items are shown.

bzang: I like it. Will make that change

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
```diff
@@ -2,71 +2,43 @@ import React, {PropTypes} from 'react';
...
     <div className={classNames(`activity-menu`, styles.activityMenu)}>
-      <div className={classNames(`activity-menu-exit`, styles.exit)}>
-        <div className={classNames(`menu-item`, styles.menuItem)}>
-          <div className={classNames(`button-container`, styles.buttonContainer)}>
-            <div className={classNames(`button-item`, styles.buttonItem)}>
-              <Button iconType={ICON_TYPE_EXIT} onClick={handleExit} title={`Exit`} />
-            </div>
-          </div>
+      {
+        onExit &&
```
adamweeks: `onExit` is required, do we need this?

packages/node_modules/@ciscospark/react-component-activity-menu/src/styles.css
```diff
@@ -1,92 +1,29 @@
 .activity-menu {
-  width: 280px;
+  position: relative;
```
adamweeks: Good idea here.

packages/node_modules/@ciscospark/react-component-button-controls/src/index.js
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div className={classNames(`controls-container`, styles.controlContainer)}>
+      {buttonsRendered}
+    </div>
+  );
+}
+
+ButtonControls.propTypes = {
+  buttons: PropTypes.arrayOf(PropTypes.shape({
+    label: PropTypes.string,
```
adamweeks: Would it be beneficial to export a function that properly created a button object?

packages/node_modules/@ciscospark/react-component-title-bar/src/styles.css
```diff
@@ -7,7 +7,6 @@
 
 .title {
   margin: 0;
-  padding-top: 10px;
```
adamweeks: Does removing this still get the title to align with the top of the avatar? Was basing it off of this mock: https://app.zeplin.io/project.html#pid=5891363f1e71c1f114ec20a8&sid=5899302750e270570fe1d37f

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -70,6 +87,7 @@ export function bindCallActions(dispatch, call) {
     localMediaStreamUrl: call.localMediaStreamUrl
   };
   call.on(`connected`, () => dispatch(callConnected(call)));
```
adamweeks: I'm guessing we can change these to `call.once` methods since they shouldn't be called with a `connected` or `disconnected` more than once. That will remove the watcher after it happens too. @ianwremmel right?

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -90,25 +108,28 @@ export function bindMediaActions(dispatch, call) {
...
-  // call.on(`change:localVideoDirection`, () => dispatch(updateMediaState(call)));
-  // call.on(`change:remoteAudioDirection`, () => dispatch(updateMediaState(call)));
-  // call.on(`change:remoteVideoDirection`, () => dispatch(updateMediaState(call)));
 }
 
-
 export function callUser({userId, constraints = {}}, spark) {
   return (dispatch) => {
-    dispatch(updateStatus({isCalling: true}));
+    dispatch(updateStatus({isDialing: true}));
```
adamweeks: I like the name change

packages/node_modules/@ciscospark/widget-meet/src/components/call-active/index.js
adamweeks: `thing = true && "value";` always seems like black magic to me.

packages/node_modules/@ciscospark/widget-meet/src/components/call-controls/index.js
```diff
@@ -0,0 +1,65 @@
...
+  onHangupClick,
+  showAnswerButton,
+  showCallButton,
+  showHangupButton,
+  showLabels
+}) {
+  const {formatMessage} = intl;
+  const buttons = [];
+
+  if (showAnswerButton) {
```
adamweeks: Thanks for putting this in for me ☎️

packages/node_modules/@ciscospark/widget-message-meet/src/container.js
```diff
@@ -121,16 +122,16 @@ export class MessageMeetWidget extends Component {
         <div className={classNames(`widget-message-meet`, styles.widgetMessageMeet)}>
           <div className={classNames(`banner`, styles.banner)} />
           <div className={classNames(`widget-message-meet-inner`, styles.widgetMessageMeetInner)}>
-            <div className={classNames(`activity-menu-wrapper`, styles.activityMenuWrapper)}>
-              {
-                showActivityMenu &&
+            {
+              showActivityMenu &&
+              <div className={classNames(`activity-menu-wrapper`, styles.activityMenuWrapper)}>
```
adamweeks: Good call on hiding the div.

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
bzang: Oops, missed the required field. I think the best approach is to remove the isRequired, so that  in a case where the exit button isn't needed, we don't have to show it.

packages/node_modules/@ciscospark/react-component-button-controls/src/index.js
bzang: Like a `buildButton({label, ...button})` method? I'm not sure it's necessary since these are the base props for a normal button anyways?

packages/node_modules/@ciscospark/react-component-title-bar/src/styles.css
bzang: Yea, I made that area flex boxy. When we add presence it should auto align correctly. I took it out because without a presence indicator the name sat above the midline.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
bzang: Good call, yup, these should probably be `.once` since we really don't want those actions to fire multiple times.

packages/node_modules/@ciscospark/react-component-button-controls/src/index.js
adamweeks: Yea, just something to help with providing some IDE feedback when creating these elements. Not necessary though.

packages/node_modules/@ciscospark/react-component-activity-menu/src/index.js
adamweeks: True, we may not want the ability to exit the menu.

bzang:
Tests are passing, but CircleCI is having some github issues

PR #42, opened 2017-02-24
feat: incoming calls

adamweeks:


packages/node_modules/@ciscospark/redux-module-media/src/reducer.js
```diff
@@ -55,7 +74,15 @@ export default function reducer(state = initialState, action) {
       .setIn([`remoteStream`, `url`], call.remoteMediaStreamUrl);
   }
   case CALL_DISCONNECTED: {
-    return initialState;
+    const isListening = state.getIn([`status`, `isListening`]);
+    const incomingAcknowledged = state.getIn([`incomingCall`, `hasAcknowledged`]);
+    const incomingStatus = state.getIn([`status`, `isRinging`]);
+    const incomingCall = state.getIn([`incomingCall`, `call`]);
+    return initialState
```
adamweeks: Just in case a call was ringing us when the disconnect happens, don't get rid of it.

packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -34,7 +37,7 @@ export class MeetWidget extends Component {
 
   shouldComponentUpdate(nextProps) {
     const props = this.props;
-    return nextProps.media !== props.media;
+    return nextProps.media !== props.media || nextProps.toPersonAvatar !== props.toPersonAvatar;
```
adamweeks: Sometimes the avatars wouldn't have loaded and we wouldn't get an update once they did.

packages/node_modules/@ciscospark/react-container-notifications/src/selectors.js
```diff
@@ -1,20 +1,34 @@
...
+            username: notificationActivity.actor.displayName,
+            message: notificationActivity.object.displayName,
+            avatar: avatars[notificationActivity.actor.id]
+          });
+        }
+        else if (notification.type === NOTIFICATION_TYPE_CALL) {
+
+          return Object.assign({}, notification, {
+            username: incomingCall.from.person.name,
+            message: `Incoming Call`,
```
adamweeks: Could not come up with a good way to i18n this string. Since it is in the selector, I couldn't use react-intl. Any suggestions?

packages/node_modules/@ciscospark/widget-meet/src/container.js
bzang: good catch

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -133,3 +153,40 @@ export function hangupCall(call) {
     });
   };
 }
+
+export function listeningForIncomingCalls() {
```
bzang: for consistency maybe this should be `listenForIncomingCalls`?

packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -110,19 +130,39 @@ export class MeetWidget extends Component {
...
-MeetWidget.propTypes = {
+const injectedPropTypes = {
   intl: intlShape.isRequired,
+  spark: PropTypes.object,
+  media: PropTypes.object,
+  acceptIncomingCall: PropTypes.func,
+  callUser: PropTypes.func,
+  declineIncomingCall: PropTypes.func,
+  hangupCall: PropTypes.func
+};
```
bzang: Should we set the injectedPropTypes to `.isRequired`?

packages/node_modules/@ciscospark/react-container-call-incoming/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+
+import {
+  NOTIFICATION_TYPE_CALL,
+  createNotification
+} from '@ciscospark/react-container-notifications';
+
+import Ringtone from './media/Ringtone.mp3';
+
+
+export class IncomingCall extends Component {
```
bzang: I don't know why, but for some reason I don't like creating React components that don't render elements. That just seems wrong to me since React was really made to render DOM elements efficiently. I could see most of this as a redux module as most of this is state management.
The ringtone does make sense being it's own component. Can that be pulled out into it's own presentation component so that it can be reused?

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
bzang: Should this be `listenForIncomingCalls` to make it consistent with our other action names/

packages/node_modules/@ciscospark/react-container-notifications/src/selectors.js
adamweeks: Yea, I think a good way to refactor it would be to have the components generating the notifications be in charge of creating the notification strings.

packages/node_modules/@ciscospark/react-container-call-incoming/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+      loop: true
+    });
+  }
+
+  componentDidMount() {
+    const {media, spark, onEvent} = this.props;
+    if (media.getIn([`status`, `isListening`])) {
+      return;
+    }
+    spark.phone.on(`call:incoming`, (incomingCall) => {
```
adamweeks: I actually started off with it in the redux module but changed it over to this after we needed to get current store data regarding media state when a call came in. Take a look at https://github.com/adamweeks/react-ciscospark-1/blob/6ea6a9037577c16ef483c6583a90007d314602b8/packages/node_modules/%40ciscospark/react-container-call-incoming/src/index.js#L86 to see what I'm talking about. 

I just came across the `getState` ability of thunk though. So maybe I can switch over to it. https://github.com/gaearon/redux-thunk#motivation

packages/node_modules/@ciscospark/widget-meet/src/container.js
adamweeks: Sure

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
adamweeks: Good call

packages/node_modules/@ciscospark/react-container-call-incoming/src/index.js
adamweeks: Yea, I think I'll split the sounds out into their own thing, then create a module for the sounds.

adamweeks:
@bzang ready for another round. Pretty much refactored everything based off your feedback 😄

PR #43, opened 2017-02-27
feat(widget-message-meet): implement initial activity feature

adamweeks:


packages/node_modules/@ciscospark/widget-message-meet/src/reducer.js
```diff
@@ -16,6 +17,9 @@ export const initialState = fromJS({
 
 export function reducer(state = initialState, action) {
   switch (action.type) {
+  case SHOW_ACTIVITY_MENU:
+    return state.set(`showActivityMenu`, true);
```
bzang: Is there a better name for this State? Having the action and the actual state both be named `showActivityMenu` seems a bit confusing. This pattern exists elsewhere, so maybe we won't want to change it now, but lets come up with a pattern that we can use for later stuff and when we go back and refactor. Some not so great names I've come up with are `hasActivityMenu`, `isVisibleActivityMenu`, and the ever popular `isShowingActivityMenu`.

adamweeks:
Ok @bzang I simplified the actions a bit, could you give it a quick re-check?

PR #46, opened 2017-02-28
feat(widget-meet): Video audio mute

bzang:
Add controls for audio and video mute. Also cleans up styles and fixes some bugs found along the way.


packages/node_modules/@ciscospark/widget-meet/src/components/call-inactive/index.js
```diff
@@ -3,41 +3,47 @@ import {intlShape} from 'react-intl';
...
-        />
+        <ButtonControls buttons={buttons} showLabels />
       </div>
     </div>
   );
 }
 
 InactiveCall.propTypes = {
-  avatarImage: PropTypes.object,
+  toPersonAvatar: PropTypes.object,
```
adamweeks: Isn't the avatar just a string?

packages/node_modules/@ciscospark/widget-meet/src/components/call-inactive/styles.css
```diff
@@ -2,6 +2,8 @@
   position: relative;
   flex: 1 1 auto;
   text-align: center;
+  background: transparent url('../../media-bg.jpg') no-repeat;
```
adamweeks: Cool, I'll be able to use this with the incoming call too.

packages/node_modules/@ciscospark/widget-message-meet/src/container.js
```diff
@@ -196,7 +196,7 @@ const injectedPropTypes = {
   fetchToUser: PropTypes.func.isRequired,
   setActivityMenuVisible: PropTypes.func.isRequired,
   toggleActivityMenu: PropTypes.func.isRequired,
-  spark: PropTypes.object.isRequired,
+  spark: PropTypes.object,
```
adamweeks: Why remove this? Was there a point where it doesn't get injected in time?

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -54,10 +54,7 @@ import {
 import styles from './styles.css';
 import messages from './messages';
 
-import {
-  EVENT_MESSAGES_CREATED,
-  EVENT_MESSAGES_UNREAD
-} from './constants';
+import events from './constants';
```
adamweeks: I like it!

packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -70,10 +70,34 @@ export class MeetWidget extends Component {
     props.hangupCall(media.get(`call`));
   }
 
+  @autobind
+  handleStartSendingAudio() {
+    this.props.media.get(`call`).startSendingAudio();
```
adamweeks: I have mixed feelings about doing these actions this way and not using redux actions. I guess since we have all of the media event listeners setup, it doesn't really matter.

packages/node_modules/@ciscospark/widget-message-meet/src/container.js
bzang: Yea when the component first loads spark isn't guaranteed to be present so we actually get a propType warning.

packages/node_modules/@ciscospark/widget-meet/src/container.js
bzang: Yea... I went back and forth on this one. Originally I was having redux thunks do this, but if we use the event listeners then a double event fire would happen. We unfortunately can't get rid of the events because really the `call` object is the single source of truth.

packages/node_modules/@ciscospark/widget-meet/src/components/call-inactive/index.js
bzang: Good catch

packages/node_modules/@ciscospark/widget-message-meet/src/container.js
adamweeks: Gotcha

PR #47, opened 2017-02-28
fix(react-container-activity-list): Add whitelist for activity types

bzang:
Fix issue where we create blank activity items in the activity list for items where we don't have rendering built.

### DIFF
--- a/packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
+++ b/packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
@@ -12,6 +12,13 @@ const getFlags = (state) => state.flags.flags;
 const getInFlightActivities = (state) => state.activity.get(`inFlightActivities`).toArray();
 const getLastAcknowledgedActivityId = (state) => state.conversation.lastAcknowledgedActivityId;
 
+export const visibleActivityTypes = [
+  `tombstone`,
+  `post`,
+  `share`,
+  `create`
+];
+
 /**
  * This loops through our conversation activities and computes an array
  * of 'visible activities' to be used with the ActivityList component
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/redux-module-conversation/src/actions.js
```diff
@@ -114,7 +114,7 @@ export function createConversationWithUser(userId, spark) {
       participants: [userId]
     }, {
       latestActivity: true,
-      activitiesLimit: 30,
+      activitiesLimit: 100,
```
bzang: Needed to up the limit because if `updates` activity items are the only ones available, there is no way to scroll up to trigger the list to load more history.

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -25,56 +32,58 @@ export const getActivityList = createSelector(
     let shouldDisplayNewMessageMarker = false;
 
     activities.forEach((activity) => {
-      // Insert day separator if this activity and last one happen on a different day
-      const activityMoment = moment(activity.published, moment.ISO_8601);
-      const activityDay = activityMoment.endOf(`day`);
-      const sameDay = activityDay.diff(lastDay, `days`) === 0;
-      if (lastDay && !sameDay) {
+      if (visibleActivityTypes.includes(activity.verb)) {
```
adamweeks: Nice!

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -25,56 +32,58 @@ export const getActivityList = createSelector(
...
+          visibleActivityList.push({
+            type: ITEM_TYPE_NEW_MESSAGE_SEPARATOR,
+            key: `new-messages-${activity.id}`
+          });
+          shouldDisplayNewMessageMarker = false;
+        }
+
+        // Actual visible activity item
+        // additional items don't repeat user avatar and name
+        const isAdditional = sameDay && lastActorId === activity.actor.id && lastVerb === activity.verb;
```
adamweeks: We'll need to eventually match the other clients on what is actually considered an "additional" item. I think it is only within a certain amount of minutes.

Not like you introduced this issue, just jogged my memory.

packages/node_modules/@ciscospark/redux-module-conversation/src/actions.js
adamweeks: Haha, I got that issue too with all my call testing.

PR #54, opened 2017-03-02
feat(widget-meet): Draggable self view

bzang:
Add ability to drag self view around active video.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
+++ b/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
@@ -169,7 +169,8 @@ export function listenForIncomingCalls(spark, locusUrl, onEvent) {
           incomingCall.once(`disconnected`, removeIncoming);
           // Once we are connected, our connected call state will handle disconnects
           incomingCall.once(`connected`, () => incomingCall.off(`disconnected`, removeIncoming));
-
+          bindCallActions(dispatch, incomingCall);
+          bindMediaActions(dispatch, incomingCall);
           // Emit the event
           onEvent(`calls:ringing`, {
             actor: incomingCall.from.person,
@@ -201,12 +202,10 @@ export function acceptIncomingCall(incomingCall, spark) {
   return (dispatch) => {
     spark.phone.createLocalMediaStream()
       .then((localMediaStream) => {
+        dispatch(storeCall(incomingCall));
         incomingCall.answer({localMediaStream})
           .then(() => {
-            dispatch(callConnected(incomingCall));
             dispatch(removeCallIncoming(incomingCall));
-            bindCallActions(dispatch, incomingCall);
-            bindMediaActions(dispatch, incomingCall);
           });
       });
   };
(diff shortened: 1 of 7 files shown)


packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -169,7 +169,8 @@ export function listenForIncomingCalls(spark, locusUrl, onEvent) {
           incomingCall.once(`disconnected`, removeIncoming);
           // Once we are connected, our connected call state will handle disconnects
           incomingCall.once(`connected`, () => incomingCall.off(`disconnected`, removeIncoming));
-
+          bindCallActions(dispatch, incomingCall);
```
adamweeks: Good catch

packages/node_modules/@ciscospark/widget-meet/src/components/call-active/index.js
```diff
@@ -62,9 +65,16 @@ export default function ActiveCall({
       </div>
       {
         localMediaStreamUrl && sendingVideo &&
-        <div className={classNames(styles.localVideo, `local-video`)}>
-          <Video audio={false} src={localMediaStreamUrl} />
-        </div>
+        <Draggable
```
adamweeks: This is cool

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -201,12 +202,10 @@ export function acceptIncomingCall(incomingCall, spark) {
   return (dispatch) => {
     spark.phone.createLocalMediaStream()
       .then((localMediaStream) => {
+        dispatch(storeCall(incomingCall));
```
adamweeks: The reason I didn't store the call here when I initially wrote it was because I wanted to make sure the `.answer` succeeded before storing. We'll need make sure to do more work around handling failing cases of all call statuses.

packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -64,7 +68,9 @@ export class MeetWidget extends Component {
 
   shouldComponentUpdate(nextProps) {
     const props = this.props;
-    return nextProps.media !== props.media || nextProps.toPersonAvatar !== props.toPersonAvatar;
+    return nextProps.media !== props.media
```
adamweeks: Surprised eslint allowed you to do this without an ignore!

bzang: I may have added the rules into our local eslint rules... It just seemed a bit silly to not allow multiple lines here.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
bzang: Gotcha, yea, we definitely need more error handling. I'm hoping that we get designs for that soon so we can build out the UI for it.

PR #55, opened 2017-03-02
feat(widget-message-meet): Add ringtones, and call duration

bzang:
Most of this is in: https://github.com/ciscospark/react-ciscospark/pull/54
Only thing relevant are the last 3 commits.


adamweeks:
You've got a snapshot failing on this.

packages/node_modules/@ciscospark/react-component-ringtone/src/index.js
```diff
@@ -3,7 +3,7 @@ import {Component, PropTypes} from 'react';
 import {Howl} from 'howler';
 
 import RingtoneIncoming from './media/Ringtone.mp3';
-import RingtoneRingback from './media/Ringback_v7_norm.wav';
+import RingtoneRingback from './media/Ringback_v2.mp3';
```
adamweeks: Did you convert this or get it elsewhere?

packages/node_modules/@ciscospark/widget-message-meet/src/container.js
```diff
@@ -178,7 +182,14 @@ export class MessageMeetWidget extends Component {
                 image={toUserAvatarUrl}
                 name={toUser.displayName}
                 onMenuClick={this.handleMenuClick}
-              />
+              >
+                {
+                  callStartTime &&
+                  <div className={classNames(`title-bar-call-timer`, styles.callTimer)}>
```
adamweeks: We should also figure out a way to make this available in the `widget-meet` so the call timer would be available for anyone just using that widget.

bzang: yea. The timer component can be used wherever and has an internal store (which I really don't like, but makes a lot of sense here). The designs specifically don't include the duration anywhere, but I do think we should have that somewhere in `widget-meet`.

packages/node_modules/@ciscospark/react-component-ringtone/src/index.js
bzang: got it from the web client. It's 1/5 the size of the wav so I figure it's better than delivering a 750kb file to the user.

adamweeks: Cool, we can probably remove the `.wav` rules from webpack then.

PR #56, opened 2017-03-03
feat: call history in activity list

adamweeks:


packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
```diff
@@ -48,6 +48,7 @@ export default function ActivityItem(props) {
 ActivityItem.propTypes = {
   activity: PropTypes.object,
   avatarUrl: PropTypes.string,
+  callData: PropTypes.object,
```
bzang: Feels kinda weird that we're passing call specific stuff into this component. Would it make more sense to construct a single `activity` object at the selector level?

packages/node_modules/@ciscospark/react-component-activity-system-message/src/index.js
```diff
@@ -8,32 +8,44 @@ import styles from './styles.css';
...
       systemMessage = <FormattedMessage {...messages.someoneDelete} values={{name}} />;
     }
+    break;
+  }
+  case SYSTEM_MESSAGE_VERB_UPDATE: {
+    systemMessage = <FormattedMessage {...messages[callData.status]} values={{...callData.callInfo}} />;
+    break;
+  }
+  default:
+    return null;
```
bzang: What do you think about having a ticket to move these messages out of here? Having the actual strings and `intl` in here just feels dirty to me.

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -105,3 +118,72 @@ export const getActivityList = createSelector(
...
+    return false;
+  }
+  const verb = VISIBLE_ACTIVITY_VERBS[activity.verb];
+  if (verb.objectTypes) {
+    if (verb.objectTypes.indexOf(activity.object.objectType) === -1) {
+      return false;
+    }
+  }
+  return true;
+}
```
bzang: Nice. Pulling this logic out makes this whole thing so much cleaner

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -272,21 +270,19 @@ export class MessageWidget extends Component {
...
       const activity = event.data.activity;
-      const isVisibleContent = VISIBLE_ACTIVITY_VERBS.indexOf(activity.verb) !== -1
-        && VISIBLE_ACTIVITY_TYPES.indexOf(activity.object.objectType) !== -1;
       // Ignore activity from other conversations
       if (activity.target.id === conversationId) {
-        if (isVisibleContent) {
+        if (activity.object.objectType === `activity`) {
+          props.receiveMercuryActivity(activity);
+        }
+        else {
```
bzang: oooo nice! Turns out if you filter things before hand you don't have to check it over and over again!

packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
adamweeks: Yea, I think that makes a lot of sense. I'll change that up.

packages/node_modules/@ciscospark/react-component-activity-system-message/src/index.js
adamweeks: It would be great to be able to do this at the selector and just pass it as a param, but I still can't figure out a good way to get i18n in the selector.

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
adamweeks: Yea, we were doing something similar to this in a few different places in the code base.

packages/node_modules/@ciscospark/react-component-activity-system-message/src/index.js
bzang: Yea, only thing I could think of is passing an additional param into the selector that's an `errorMessages` object or something and do the `formatMessage` part in the main widget, but that seems really hacky.

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -75,20 +75,28 @@ export const getActivityList = createSelector(
...
+        if (activity.verb === `update`) {
+          callData = getCallData(activity, currentUser);
+        }
+
         const visibleActivity = {
           type: ITEM_TYPE_ACTIVITY,
-          activity,
+          activity: {
+            ...activity,
+            callData
```
adamweeks: `callData` is added to the activity object so we don't clutter up the top level params.

adamweeks:
@bzang ready for round two

packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
```diff
@@ -46,9 +46,21 @@ export default function ActivityItem(props) {
...
+  }),
   avatarUrl: PropTypes.string,
-  callData: PropTypes.object,
+  callData: PropTypes.shape({
+    callInfo: PropTypes.shape({
+      duration: PropTypes.string,
+      name: PropTypes.string
+    }),
+    status: PropTypes.string
+  }),
```
bzang: Did you mean to have the `callData` as it's own prop? I was thinking this would be inside the `activity` object instead being it's own separate prop because the activity object has all the rest of the information. What do you think?

packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
```diff
@@ -46,9 +46,21 @@ export default function ActivityItem(props) {
...
 /* eslint-disable-reason passing props via destructuring */
 ActivityItem.propTypes = {
-  activity: PropTypes.object,
+  activity: PropTypes.shape({
+    content: PropTypes.string,
+    displayName: PropTypes.string,
+    files: PropTypes.shape({
+      items: PropTypes.array
+    })
+  }),
```
bzang: nice on the proptype shapes

packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
adamweeks: Yea, I left it at its own level here because it needed to be split from activity at some point since the system message doesn't support an activity object. If it gets to where it is too confusing, we can merge them back in and split it out of the activity object further down the chain.

Issue #53, opened 2017-03-02
Typing indicator doesn't remove after sending


adamweeks:
We are not sending a stop typing status along with our message, so it appears that the user is typing even after sending.

PR #59, opened 2017-03-06
fix(redux-module-activity): clear typing indicator when sending activity

adamweeks:
Went ahead and prefixed the activity module as well.


adamweeks:
Closes #53

PR #63, opened 2017-03-08
fix(widget-message-meet): Start conversation with person with no account 

bzang:
Fix issue where we were not loading widget when toPerson does not have an account.


packages/node_modules/@ciscospark/react-component-video/src/index.js
```diff
@@ -3,13 +3,19 @@ import classNames from 'classnames';
...
 
-export default function Video({src}) {
+export default function Video({src, audio}) {
   // Need autoPlay to start the video automagically
   return (
-    <video autoPlay className={classNames(styles.video)} src={src} />
+    <video
+      autoPlay
+      className={classNames(styles.video)}
+      muted={audio}
```
ianwremmel: maybe `audioMuted` instead of just `audio`? i would expect `audio === true` to produce audio, not mute audio.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -34,40 +34,45 @@ class SparkComponent extends Component {
       props.updateSparkStatus({authenticating: spark.isAuthenticating});
     });
 
-    if (spark.mercury) {
+    if (spark.mercury && !spark.mercury.connected) {
```
ianwremmel: This section looks right, but take maybe take a look at both the example-phone app and the web client's mercury handling just in case.

PR #64, opened 2017-03-08
styles(widget-message-meet): Add docs and styles for browser globals

bzang:
Not 100% sure if this is the right way to do it, would love some thoughts.


packages/node_modules/@ciscospark/widget-message-meet/README.md
```diff
@@ -132,11 +134,13 @@ Create a container where you would like to embed the widget and add the followin
 
 If you need additional behaviors or need to do additional work before the widget loads, it may be useful for to programmatically instatiate the widget after the intial page loads.
 
+Be sure to add the `ciscospark-widget` class to your div to ensure that appropriate styles are applied.
+
 ```html
-<div id="ciscospark-widget" />
+<div id="ciscospark-widget-id" class="ciscospark-widget"/>
```
ianwremmel: maybe use something more generic for the id name so you don't have to add "id" to differentiate it from the class name.

packages/node_modules/@ciscospark/widget-message-meet/README.md
```diff
@@ -132,11 +134,13 @@ Create a container where you would like to embed the widget and add the followin
 
 If you need additional behaviors or need to do additional work before the widget loads, it may be useful for to programmatically instatiate the widget after the intial page loads.
 
+Be sure to add the `ciscospark-widget` class to your div to ensure that appropriate styles are applied.
```
ianwremmel: Could `ciscospark.widget()` add the class?

bzang: hmmm yea, that might work

packages/node_modules/@ciscospark/widget-message-meet/README.md
bzang: good call

PR #78, opened 2017-03-16
feat(widget-example): Example app that uses new base

bzang:
Looking for some feedback. Not Ready Yet
** DO NOT MERGE **


adamweeks:
I like the separation on the `widget-base` but I'm not sure I understand the purpose of the example.

bzang:
The example is just a shell to test out how it would look to use the widget-base and see if it makes sense or is too confusing. Wanted to get that sorted out before attempting to switch a real widget over.

adamweeks:
Ah, ok. Since it was in packages I was thinking someone might actually use it. Looks reasonable to me

packages/node_modules/@ciscospark/widget-base/src/setup-data-api.js
```diff
@@ -0,0 +1,32 @@
+import _ from 'lodash';
```
ianwremmel: does `babel-plugin-lodash` work if you use `_.camelCase()` rather than `import {camelCase} from 'lodash'`

adamweeks:
@bzang can we close this?

bzang:
Yea, gonna close this... once I have something ready to review again with all the recent changes, I'll open a new one.

PR #79, opened 2017-03-16
feat(redux-module-mercury): move mercury status out of spark module

adamweeks:
This makes our redux-spark module depend on the mercury module, but it sets things up so I can do the reconnect logic.


packages/node_modules/@ciscospark/redux-module-mercury/src/index.js
```diff
@@ -0,0 +1,51 @@
+import {Map} from 'immutable';
+
+export const SET_CONNECTING = `mercury/SET_CONNECTING`;
+export const SET_DISCONNECTING = `mercury/SET_DISCONNECTING`;
+export const SET_CONNECTED = `mercury/SET_CONNECTED`;
+export const SET_DISCONNECTED = `mercury/SET_DISCONNECTED`;
```
bzang: Doesn't look like these disconnect action names aren't being used anywhere? Are these gonna be used when everything gets migrated over?

packages/node_modules/@ciscospark/redux-module-mercury/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const initialState = new Map({
+  connected: false,
+  connecting: false
+});
```
bzang: Can we set these into a status object to be consistent with the spark module?

packages/node_modules/@ciscospark/redux-module-mercury/src/index.js
adamweeks: Oops, leftovers.

packages/node_modules/@ciscospark/redux-module-mercury/src/index.js
adamweeks: Sure.

adamweeks:
@bzang changes requested have been added

PR #83, opened 2017-03-22
tooling: Initial framework for /clientmetrics

bzang:
Initial framework for /clientmetrics


packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
```diff
@@ -36,6 +36,7 @@ export function sendPerfMetric({key, data}, spark) {
       hostname: window.location.hostname,
       pathname: window.location.pathname,
       widgetVersion: process.env.REACT_CISCOSPARK_VERSION,
+      jsSdkVersion: spark.version,
```
ianwremmel: this should be done automatically by the sdk, probably [here](https://github.com/ciscospark/spark-js-sdk/blob/master/packages/plugin-metrics/src/client-metrics-batcher.js#L12-L15)

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
```diff
@@ -0,0 +1,116 @@
+const METRIC_NS = `ciscospark-widget-`;
+const LOAD_START = `startLoad`;
+
+/**
+ * Temporarily save metric to sessionStorage
+ * @param {string} key
+ * @param  {any} data
+ * @returns {null}
+ */
+export function saveTempMetric(key, data) {
```
ianwremmel: Let's get these primitives into the sdk

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
```diff
@@ -0,0 +1,116 @@
...
+  window.sessionStorage.setItem(key, data);
```
ianwremmel: you could use a map just as easily as session storage

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
```diff
@@ -25,14 +25,21 @@ export function getTempMetric(key) {
...
-    protocol: window.location.protocol,
-    hostname: window.location.hostname,
-    pathname: window.location.pathname,
-    widgetVersion: process.env.REACT_CISCOSPARK_VERSION,
-    data
-  });
-  return spark.metrics.sendUnstructured(`${METRIC_NS}${key}`, formattedData);
+  const [name, action] = key.split(`:`);
+  const formattedData = {
+    type: [`operational`],
```
ianwremmel: we should add sdk primitives for sendbehavioral, sendOperational, etc

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
bzang: Yea, we should probably do that there, but in efforts to get this out into the wild before the weekend, can you create a ticket to do that in the SDK in the next sprint?

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
bzang: Yea, probably a better idea... also easier to deal with in unit tests. I don't have to mock out sessionStorage.

bzang:
@ianwremmel, ready for another round. I'm gonna leave the calls in here for now since we want to get this out asap. Let's add the helper methods into the sdk after we get through this push.

packages/node_modules/@ciscospark/react-redux-spark/src/actions.js
```diff
@@ -21,8 +23,15 @@ export function storeSparkInstance(spark) {
...
+    metrics.saveElapsedTime(`registerDevice:start`);
+    metrics.sendTempMetric(`registerDevice:start`, spark);
     dispatch(updateSparkStatus({registering: true}));
     return spark.device.register()
-      .then(() => dispatch(updateSparkStatus({registering: false, registered: true})));
+      .then(() => {
+        metrics.saveElapsedTime(`registerDevice:end`);
+        metrics.sendTempMetric(`registerDevice:end`, spark);
+        metrics.sendDurationMetric(`registerDevice:duration`, `registerDevice:start`, `registerDevice:end`, spark);
+        dispatch(updateSparkStatus({registering: false, registered: true}));
```
adamweeks: We could `return dispatch` here to keep it thunkable (?), but I don't think it is needed.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -24,10 +26,13 @@ class SparkComponent extends Component {
 
     if (!spark) {
       spark = createSpark(accessToken);
+      metrics.sendLoadStart(spark);
+      metrics.sendElapsedTimeMetric(`sparkObject:loaded`, spark);
       props.storeSparkInstance(spark);
     }
 
     spark.listenToAndRun(spark, `change:canAuthorize`, () => {
+      metrics.sendElapsedTimeMetric(`sparkObject:authenticated`, spark);
```
adamweeks: Do we only want to fire this once `spark.canAuthorize` is true? And only fire it once per session?

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
```diff
@@ -0,0 +1,126 @@
...
+ * @returns {Promise}
+ */
+export function sendLoadStart(spark) {
+  const data = getTempMetric(LOAD_START);
+  return sendPerfMetric(data, spark);
+}
+
+/**
+ * Save start time
+ * @param  {object} spark
```
adamweeks: This param isn't used

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
bzang: ooo i think we should actually capture a little more here... if this hits multiple times during a single session then we can actually see that there are multiple connects.

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
bzang: good catch

packages/node_modules/@ciscospark/react-redux-spark/src/actions.js
bzang: yea we can do that just in chase it's needed.

adamweeks:
Should we do metrics on the actual initial rendering as well? Maybe in the componentWillMount/DidMount or something?

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
adamweeks: True, maybe the metric should be `sparkObject:authenticationChanged`?

bzang:
Yea, @adamweeks , I think didMount is a good idea.

packages/node_modules/@ciscospark/redux-module-mercury/src/index.js
```diff
@@ -40,12 +41,18 @@ function updateStatusConnected(value) {
...
   return (dispatch) => {
+    metrics.saveElapsedTime(`mercuryConnection:start`);
+    metrics.sendTempMetric(`mercuryConnection:start`, spark);
     spark.mercury.on(`change:connecting`, () => {
       dispatch(updateStatusConnecting(spark.mercury.connecting));
     });
     spark.mercury.on(`change:connected`, () => {
+      metrics.saveElapsedTime(`mercuryConnection:end`);
+      metrics.sendTempMetric(`mercuryConnection:end`, spark);
+      metrics.sendDurationMetric(`mercuryConnection:duration`, `mercuryConnection:start`, `mercuryConnection:end`, spark);
```
adamweeks: Does it makes sense to change these to constants? They are used a few different times and I'm prone to making typing mistakes.

bzang: Where do you think we should store these? maybe along side the metrics helper?

adamweeks: Nothing that extreme, maybe just on line 50 😄

adamweeks: I'm just terrified of using a string more than once 🤣

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
```diff
@@ -96,7 +95,7 @@ function getElapsedTime() {
 
 /**
  * Save time elapsed for event
- * @type {[type]}
+ * @type {type}
```
ianwremmel: looks like this comment block just generated completely incorrectly. probably because of the gap at line 100. you'll want `@param` and `@returns`, not `@type`

PR #96, opened 2017-03-28
fix(redux-module-media): Fix calling not recieving remote media

bzang:
This needs one more change in `package.json` whenever the fixes make it into the `sdk`


packages/node_modules/@ciscospark/redux-module-media/src/reducer.js
```diff
@@ -52,7 +45,16 @@ export default function reducer(state = initialState, action) {
   case UPDATE_STATUS:
     return state.mergeIn([`status`], action.payload.status);
   case UPDATE_CALL_STATE: {
+    let remoteAudioMuted, remoteVideoMuted;
     const call = action.payload.call;
+    try {
+      remoteAudioMuted = call.remoteAudioMuted;
```
ianwremmel: can this ever throw?

bzang: Yup, turns out `change` events get fired on the `call` object before `call.remoteAudioMuted` and `call.remoteVideoMuted` are available as derived values, so we get an Uncaught Type Error. There should be a fix for the SDK, but for now this is a workaround.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -140,7 +140,10 @@ export function callUser({userIdOrEmail, constraints = {audio: true, video: true
       .then((localMediaStream) => {
         const call = spark.phone.dial(
           userIdOrEmail,
-          Object.assign({offerOptions: {offerToReceiveVideo: true, offerToReceiveAudio: true}}, {localMediaStream})
+          Object.assign({
+            constraints,
```
adamweeks: I thought constraints weren't needed when passing a local media stream?

bzang: Yea that's what we thought, but parts of the rest of the webrtc and calling code expects it for state management. So when it isn't set change events don't fire correctly.

PR #97, opened 2017-03-28
Performance Enhancements

bzang:
Did a lot to parallelize some calls and get rid of bad things that were making us redraw.
This gets my manual testing down to about 3.5s consistently. Will need to see how metrics look and how it looks in production.


packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -6,11 +6,11 @@ import _ from 'lodash';
 
 import {ITEM_TYPE_ACTIVITY, ITEM_TYPE_DAY_SEPARATOR, ITEM_TYPE_NEW_MESSAGE_SEPARATOR} from '@ciscospark/react-component-activity-list';
 
-const getActivities = (state) => state.conversation.activities.toArray();
-const getAvatars = (state) => state.avatar.get(`items`).toJS();
-const getCurrentUser = (state) => state.user.get(`currentUser`);
+const getConversationActivities = (state) => state.conversation.activities;
```
adamweeks: Did you see a big change in load times when moving the immutable deconstructors into the actual sector?

PR #98, opened 2017-03-28
feat(widget-message-meet): hangup active call on unload

adamweeks:


packages/node_modules/@ciscospark/widget-message-meet/src/container.js
```diff
@@ -110,7 +111,19 @@ export class MessageMeetWidget extends Component {
   }
 
   componentWillUnmount() {
+    // Hang up active call (if there is one)
+    const props = this.props;
+    const {
+      media
+    } = props;
```
bzang: I think with our changes we can access `this.props.media` directly now right?

adamweeks: Good call, fixed.

PR #105, opened 2017-03-30
fix(widget-message-meet): Load order fix, move convo to immutable

bzang:


packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -6,12 +6,12 @@ import _ from 'lodash';
 
 import {ITEM_TYPE_ACTIVITY, ITEM_TYPE_DAY_SEPARATOR, ITEM_TYPE_NEW_MESSAGE_SEPARATOR} from '@ciscospark/react-component-activity-list';
 
-const getConversationActivities = (state) => state.conversation.activities;
+const getConversationActivities = (state) => state.conversation.get(`activities`);
```
adamweeks: Should we just have a getConversation method and get the immutable object within the selector? Does that improve timing at all?

adamweeks: Actually, maybe not if it causes the selector to fire again

bzang: yea, tried that out and it actually slowed things down because fires on every conversation change.

packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
```diff
@@ -11,7 +11,7 @@ import {
   UPDATE_MERCURY_STATE
 } from './actions';
 
-export const initialState = {
+export const initialState = new Map({
   activities: new OrderedMap(),
   id: null,
   lastAcknowledgedActivityId: null,
```
adamweeks: Should we put all of the `isX` items into the status object to match the other reducers?

packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
```diff
@@ -65,17 +58,18 @@ export default function reducer(state = initialState, action) {
...
-      locusUrl,
-      url,
-      isFetching: false,
-      isLoaded: true,
-      participants: participants.items
-    });
+    return state
+      .set(`activities`, activities)
+      .merge({
+        defaultActivityEncryptionKeyUrl,
```
adamweeks: Should we put the conversation object that the sdk is expecting into its own property of the reducer? That way we can do `conversation.conversationObject.toJS()` when calling sdk methods. It might speed up the `toJS` methods since we won't have to convert all of the activities every time.

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
adamweeks: Yup, good to know.

packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
bzang: Yea, that's probably a good idea. Didn't want to do too much here as a "bug fix". I have that set up in the "spaces" I've been putting together.

packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
bzang: Actually i'm wondering if we should keep everything as an immutable object until it hits an sdk method and then we convert with `.toJS()`. That may actually speed stuff up.

adamweeks: I like the idea of not converting before sending. We could create a utility method that gets the items the sdk needs out of the immutable object and just have it call that method right before the sdk. That way we don't have to `.toJS` the entire conversation object, but I don't know how much effort that would be and time saved.

PR #110, opened 2017-03-31
Automation Tests

ianwremmel:


test/journeys/specs/widget-message-meet/messaging.js
```diff
@@ -0,0 +1,92 @@
...
+    // Leaves the browser open for further testing and inspection
+    after(() => {
+      browser.timeouts(`implicit`, 1200000);
+      return browser.waitUntil(() => false, 1200000, `done waiting: bye`, 10000);
+    });
+  }
+
+  let mccoy, spock;
+  process.env.CISCOSPARK_SCOPE = [
+    `webexsquare:get_conversation`,
```
adamweeks: What is this scope for?

test/journeys/specs/widget-message-meet/messaging.js
```diff
@@ -0,0 +1,92 @@
...
+      window.openWidget(localAccessToken, localToUserEmail);
+    }, spock.token.access_token, mccoy.email);
+  });
+
+  after(() => mccoy && mccoy.spark.mercury.disconnect());
+
+  it(`sends and receives messages`, () => {
+    browser.waitForVisible(`[placeholder="Send a message to ${mccoy.displayName}"]`);
+    assert.match($(`.ciscospark-system-message`).getText(), /You created this conversation/);
+    $(`[placeholder="Send a message to ${mccoy.displayName}"]`).setValue(`Oh, I am sorry, Doctor. Were we having a good time?\n`);
```
adamweeks: Does the `\n` make it actually send? I know we are waiting for a key event of enter/return before sending

ianwremmel: Yea, `\n` was the only way I could find to send the message. We should refactor the composer so it's actually a form at some point.

test/journeys/specs/widget-message-meet/messaging.js
ianwremmel: could've sworn i replied to this...this scope is the old internal scope. I'm not sure what checks it at this point, but it's what we've got in all the other tests; i don't want to get rid of it until we're absolutely certain we can.

PR #113, opened 2017-03-31
feat(react-redux-spark): Multi widgets

bzang:
With some other tweaks


packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
```diff
@@ -10,62 +10,70 @@ import Spark from '@ciscospark/spark-core';
...
-    credentials: {
-      authorization: {
-        access_token: accessToken || credentials.access_token
-      }
-    },
-    config: {
-      logger: {
-        level: process.env.NODE_ENV === `production` ? undefined : `trace`
+  // Check if we already have a spark instance in memory
+  let ciscosparkInstance = window.ciscosparkInstance;
```
ianwremmel: Would it be better to attach it somewhere to `window.ciscospark`?

packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
```diff
@@ -10,62 +10,70 @@ import Spark from '@ciscospark/spark-core';
...
+
+  if (typeof ciscosparkInstance !== `object`) {
```
ianwremmel: `(!(ciscosparkInstance instanceof SparkCore))`?

packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
bzang: I originally wanted to but didn't want to chance the instance getting overwritten by someone attaching the SDK there.

ianwremmel: gotcha. I think this works then. at some point, i'd like to consolidate how a spark global instance from the sdk can coordinate, but that's probably not necessary until the widgets can run strictly on the `ciscospark` package.

bzang:
@ianwremmel and @adamweeks , ready for round 2. Got the tests passing. Wanna get this out ASAP before Monday morning into production so we have some time to quickly tests everything in the production UI

packages/node_modules/@ciscospark/react-component-utils/src/metrics.js
```diff
@@ -45,38 +45,39 @@ export function sendMetricsInQueue(spark) {
...
-        iteration: metricsCount[key],
-        data
-      }, fields)
-    };
-    return spark.metrics.submitClientMetrics(`${METRIC_NS}${name}`, formattedData);
-  }
-  catch (e) {
-    return console.warn(`clientMetrics: ${key} metrics was not sent`, e); // eslint-disable-line no-console
-  }
+export function sendPerfMetric({key, data, fields, tags}, spark) { // eslint-disable-line no-unused-vars
```
ianwremmel: Is this commented out on purpose?

PR #114, opened 2017-04-03
feat(widget-meet): Render video srcObject instead of url

bzang:
With `createObjectURL` losing support for media streams, we've created a work around for the react video element to support a `srcObject`

### DIFF
--- a/packages/node_modules/@ciscospark/react-component-video/src/index.js
+++ b/packages/node_modules/@ciscospark/react-component-video/src/index.js
@@ -3,13 +3,21 @@ import classNames from 'classnames';
 
 import styles from './styles.css';
 
-export default function Video({src, audioMuted}) {
+export default function Video({src, srcObject, audioMuted}) {
   // Need autoPlay to start the video automagically
+
+  function getEl(el) {
+    if (el && srcObject) {
+      el.srcObject = srcObject;
+    }
+  }
+
   return (
     <video
       autoPlay
       className={classNames(styles.video)}
       muted={audioMuted}
+      ref={getEl}
       src={src}
     />
   );
@@ -17,5 +25,6 @@ export default function Video({src, audioMuted}) {
 
 Video.propTypes = {
   audioMuted: PropTypes.bool,
-  src: PropTypes.string.isRequired
+  src: PropTypes.string,
+  srcObject: PropTypes.object
 };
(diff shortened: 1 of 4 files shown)


packages/node_modules/@ciscospark/react-component-video/src/index.js
```diff
@@ -3,19 +3,28 @@ import classNames from 'classnames';
...
-export default function Video({src, audioMuted}) {
+export default function Video({src, srcObject, audioMuted}) {
   // Need autoPlay to start the video automagically
+
+  function getEl(el) {
+    if (el && srcObject) {
+      el.srcObject = srcObject;
+    }
+  }
+
```
bzang: We have to do this because React strips out all non-whitelisted props. Currently `srcObject` is not allowed

adamweeks:
We should track this as well: https://github.com/facebook/react/pull/9146

packages/node_modules/@ciscospark/react-component-video/src/index.js
```diff
@@ -3,19 +3,28 @@ import classNames from 'classnames';
...
+      ref={getEl}
       src={src}
     />
   );
 }
 
 Video.propTypes = {
   audioMuted: PropTypes.bool,
-  src: PropTypes.string.isRequired
+  src: PropTypes.string,
```
ianwremmel: Should we get rid of src entirely?

packages/node_modules/@ciscospark/redux-module-media/src/reducer.js
```diff
@@ -75,7 +77,8 @@ export default function reducer(state = initialState, action) {
     const mediaState = {
       localAudioDirection: call.localAudioDirection,
       localVideoDirection: call.localVideoDirection,
-      localMediaStreamUrl: call.localMediaStreamUrl
+      localMediaStreamUrl: call.localMediaStreamUrl,
```
ianwremmel: Should we stop using the url fields entirely?

packages/node_modules/@ciscospark/widget-meet/src/components/call-active/index.js
```diff
@@ -19,12 +19,12 @@ export default function ActiveCall({
...
   onStopSendingAudio,
   onStopSendingVideo,
-  localMediaStreamUrl,
-  remoteMediaStreamUrl,
+  localMediaStream,
+  remoteMediaStream,
   toPersonAvatar,
   toPersonName
 }) {
-  const connectedClass = remoteMediaStreamUrl && styles.callConnected;
```
adamweeks: Why are we still storing the remote/local media stream url if we're not using it?

PR #117, opened 2017-04-04
Greenkeeper/redux logger 3.0.1

adamweeks:


packages/node_modules/@ciscospark/widget-message-meet/src/store.js
```diff
@@ -13,7 +13,7 @@ const devtools = window.devToolsExtension || (() => (noop) => noop);
 const middlewares = [thunk];
 
 if (process.env.NODE_ENV !== `production`) {
-  const createLogger = require(`redux-logger`); // eslint-disable-line global-require
+  const {createLogger} = require(`redux-logger`); // eslint-disable-line global-require
```
bzang: What does the module export as `default` now?

adamweeks: Exports `logger` as default, which come with default options, but since we are creating our own, we can't use the default.

PR #123, opened 2017-04-04
chore(tooling): Build versioning

bzang:


Jenkinsfile
```diff
@@ -128,6 +122,21 @@ ansiColor('xterm') {
           }
 
           if (currentBuild.result == 'SUCCESS'){
+            stage('Bump version'){
+              sh '''#!/bin/bash -ex
+              source ~/.nvm/nvm.sh
+              nvm use v6
+              npm version patch
+              grep "version" package.json | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g' > .version
```
adamweeks: There are some good version bump npm packages that we could use if we don't want to use this magic, lol

scripts/build-packages.js
```diff
@@ -1,14 +1,16 @@
...
 
 // Run buildPackage on all of our packages
-readDir(`packages/node_modules/@ciscospark`)
-  .then((packages) =>
-    Promise.all(packages.map((pkg) => buildPackage(pkg)))
-      .catch((error) => {
-        throw new Error(error);
-      })
-  );
+console.log(new Date());
```
adamweeks: Is this supposed to be here?

Jenkinsfile
bzang: That's actually magic for Jenkins to get the version number. The only thing we're using to bump the version is 'npm version patch'

scripts/build-packages.js
bzang: Nope. Was using it to test build timing.

Jenkinsfile
```diff
@@ -128,9 +122,26 @@ ansiColor('xterm') {
           }
 
           if (currentBuild.result == 'SUCCESS'){
+            stage('Bump version'){
+              sh '''#!/bin/bash -ex
+              source ~/.nvm/nvm.sh
```
ianwremmel: At some point, we should see about either (a) replacing the executor dockerfile with something that doesn't require us to call nvm in every shell script or (b) do something like what the sdk and do all the real work in a separate container (with the way we have docker setup, there should be minimal overhead - the sdk's overhead is from running multiple containers in parallel on the same box).

package.json
```diff
@@ -94,6 +96,7 @@
     "child-process-promise": "^2.2.0",
     "css-loader": "^0.27.3",
     "denodeify": "^1.2.1",
+    "detective-es6": "^1.1.6",
```
ianwremmel: does this handle imports?

bzang: yes it does. I started out with detective on `/dist` but realized that we can't do that because our bundles bring all the code in, so there aren't actually any `require` calls.

Jenkinsfile
bzang: yea, agreed. I hate seeing redundant code like that.

PR #127, opened 2017-04-07
feat(widget-base): Widget base refactor

bzang:
New more concise widget base that moves repeated logic and hookups into a single location. This also moves the main `widget-message-meet` to the new `widget-base`. If this looks good, I'll also move the other widgets to it as well.


packages/node_modules/@ciscospark/widget-base/README.md
```diff
@@ -0,0 +1,31 @@
...
+To automatically connect reducers into a Redux store, setup Cisco Spark connections, and register your widget into the global browser store, you can call the `constructSparkWidget` method like so:
+
+``` js
+import constructSparkWidget from `@ciscospark/widget-base`;
+
+function MyWidget(props) {
+  return <div>My Widget</div>;
+}
+
+const MyWidgetWithSpark = constructSparkWidget(`nameOfMyWidget`, Widget, {
```
ianwremmel: Should this be 
```js
const MyWidgetWithSpark = constructSparkWidget(`myWidget`, MyWidget, {
```

packages/node_modules/@ciscospark/widget-base/README.md
```diff
@@ -0,0 +1,31 @@
...
+  intl: {
```
ianwremmel: Long-term, I'm not sure this is how we want to expose intl, but let's talk about it at some point. no need to hold this PR on it.

packages/node_modules/@ciscospark/widget-base/src/browser-globals.js
```diff
@@ -0,0 +1,149 @@
+import uuid from 'uuid';
+import Events from 'ampersand-events';
+
+export function setupWidgets() {
+  if (!window.ciscospark.widget || typeof window.ciscospark.widget !== `function`) {
+    window.ciscospark.widget = function widget(el) {
```
ianwremmel: Each time `setupWidgets()` gets called, `widget` will be declared as a new function. does it make sense to move widget up to the top level of this module just not export it?

packages/node_modules/@ciscospark/widget-base/src/data-api.js
```diff
@@ -0,0 +1,39 @@
+import {
+  camelCase,
+  kebabCase
```
ianwremmel: this still makes me happy everytime i see it.

packages/node_modules/@ciscospark/widget-base/src/index.js
```diff
@@ -1,134 +1,8 @@
...
-        return callback(removed);
-      }
-      return resolve(removed);
-    });
-
-    return this;
-  };
-  setupWidgets();
+if (module.hot) {
+  module.hot.accept();
```
ianwremmel: Is this enough to hot reload reducers and css?

PR #128, opened 2017-04-10
test(journeys): multi remote testing of messaging and calls

adamweeks:


adamweeks:
Contains some of #129 as well. If we get it merged first, I will update this PR.

test/journeys/specs/widget-message-meet/basic.js
```diff
@@ -23,25 +25,91 @@ describe(`Widget Message Meet`, () => {
...
   });
 
   it(`loads the user's name`, () => {
-    browser.waitUntil(() => $(`h1`).getText() !== mccoy.email);
-    assert.equal($(`h1`).getText(), mccoy.displayName);
+    browserLocal.waitUntil(() => browserLocal.getText(`h1`) !== mccoy.email);
+    assert.equal(browserLocal.getText(`h1`), mccoy.displayName);
   });
+
+  describe(`menu functionality`, () => {
```
adamweeks: Yea, we can update that.

adamweeks:
Added the ability to do tests against production (TAP) as well. These will fail until we get all of the other fixes out into production.

PR #129, opened 2017-04-10
feat(accessibility): add button aria labels

adamweeks:


packages/node_modules/@ciscospark/widget-message-meet/src/container.js
```diff
@@ -219,7 +219,7 @@ export class MessageMeetWidget extends Component {
             }
             <div className={classNames(`ciscospark-activity-menu-button-wrapper`, styles.activityMenuButtonWrapper)}>
               <div className={classNames(`ciscospark-activity-menu-button`, styles.activityMenuButton)}>
-                <Button iconType={ICON_TYPE_WAFFLE} onClick={this.handleMenuClick} />
+                <Button accessibilityLabel={`Main Menu`} iconType={ICON_TYPE_WAFFLE} onClick={this.handleMenuClick} />
```
bzang: Should the label here be capitalized? The other ones above don't seem to be.

adamweeks: The button gets its aria-label from the `accessibilityLabel` prop or just the `label` prop if there is one. Since all of our labels are capitalized, we should probably do it for the aria labels too. MDN has their example capitalized too: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

packages/node_modules/@ciscospark/widget-meet/src/components/call-active/index.js
```diff
@@ -32,16 +32,19 @@ export default function ActiveCall({
 
   const buttons = [
     {
+      accessibilityLabel: `mute audio`,
```
adamweeks: @bzang should we capitalize these?

bzang: Yea, based on the link you posted from MDN, i'd say let's capitalize the accessibility labels. I was just suggesting that we keep that consistent.

adamweeks: Cool, yea, I agree.

PR #130, opened 2017-04-11
feat(redux-module-media): hangup call on disconnect

adamweeks:
When an outgoing call was declined by the remote party, the disconnect event would not relinquish the mic and camera. This solves that issue. Will be working on a fix in the sdk as well.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
+++ b/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
@@ -128,7 +128,7 @@ export function updateWebRTCSupport(supported) {
 
 export function bindCallActions(dispatch, call) {
   call.once(`connected`, () => dispatch(callConnected(call)));
-  call.once(`disconnected`, () => dispatch(callDisconnected(call)));
+  call.once(`disconnected`, () => dispatch(hangupCall(call)));
   call.on(`change:status`, () => dispatch(updateCallState(call)));
   call.on(`change:sendingAudio`, () => dispatch(updateCallState(call)));
   call.on(`change:receivingAudio`, () => dispatch(updateCallState(call)));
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -165,6 +165,8 @@ export function callUser({userIdOrEmail, constraints = {audio: true, video: true
 
 export function hangupCall(call) {
   return (dispatch) => {
+    // Don't update call states after hangup
+    call.off();
```
adamweeks: Added this here because actually calling `hangupCall` would call itself a second time on the `disconnect` event.

bzang: Good catch

packages/node_modules/@ciscospark/redux-module-media/src/reducer.js
```diff
@@ -23,8 +23,8 @@ export const initialState = fromJS({
     call: null
   },
   callState: {
-    sendingAudio: true,
-    sendingVideo: true
+    sendingAudio: false,
+    sendingVideo: false
```
bzang: Did you test this with a slow connection to make sure thing still look right? We need to make sure that there's no chance that the buttons show up in the `false` state.

There was also a case where buttons didn't show up at all if there is any sort of connection issue during the dialing phase and the user gets stuck without a way to hangup the call.

adamweeks: We're actually looking for ```media.getIn([`callState`, `localMediaStream`]);``` when determining if the call controls display. `sendingAudio` and `sendingVideo` just control how the button is styled.

Just gave it a shot with 2G network throttling and it worked fine.

PR #131, opened 2017-04-11
fix(widget-base): Change console.warning to console.warn

bzang:
Oops


packages/node_modules/@ciscospark/widget-base/src/browser-globals.js
```diff
@@ -88,7 +88,7 @@ function setupWidgets() {
           // Otherwise create new widget store
           return createNewWidget(widgetEl);
         }
-        console.warning(`WARNING: ciscospark.widget: No HTML element provided.`); // eslint-disable-line no-console
+        console.warn(`WARNING: ciscospark.widget: No HTML element provided.`); // eslint-disable-line no-console
```
ianwremmel: Should this be an error?

bzang: I didn't think so since if you use jquery with invalid selectors it would just return undefined.

adamweeks:
@bzang LGTM?

PR #133, opened 2017-04-11
feat(react-container-activity-list): show declined outbound calls as …

adamweeks:
…unavailable


packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -164,7 +164,6 @@ function getCallData(activity, currentUser) {
   const me = _.find(callObject.participants.items, (p) => p.person.entryUUID === currentUser.id);
 
   const iDeclined = me && me.state === `DECLINED`;
-  const otherDeclined = partner && partner.state === `DECLINED`;
```
adamweeks: This status was straight from the web client, but it doesn't match with the rest of the clients.

PR #134, opened 2017-04-12
update package.json to use cross-env

moriahmaney:


package.json
```diff
@@ -4,24 +4,24 @@
   "description": "The Cisco Spark for React library allows developers to easily incorporate Spark functionality into an application.",
   "scripts": {
     "build": "npm run build:packages && npm run build:bundle",
-    "build:packages": "export NODE_ENV=${NODE_ENV:-production}; babel-node --presets=stage-1,es2015 scripts/build-packages.js",
-    "build:package": "export NODE_ENV=${NODE_ENV:-production}; babel-node --presets=stage-1,es2015 scripts/build-package.js",
-    "build:bundle": "export NODE_ENV=${NODE_ENV:-production}; babel-node --presets=stage-1,es2015 scripts/build-bundle.js",
+    "build:packages": "NODE_ENV=${NODE_ENV:-production}; babel-node --presets=stage-1,es2015 scripts/build-packages.js",
```
adamweeks: I think you have to do `cross-env NODE_ENV` here if I read the docs correctly on cross-env

package.json
```diff
@@ -81,7 +81,6 @@
     "babel-loader": "^6.2.10",
     "babel-plugin-lodash": "^3.2.11",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
-    "babel-plugin-transform-export-extensions": "^6.8.0",
```
adamweeks: What happened here?

PR #152, opened 2017-04-27
feat: support message catchup

adamweeks:


scripts/jest-package.js
```diff
@@ -7,18 +9,20 @@ import runInPackage from './utils/base-package';
  * @returns {Promise}
  */
 export default function jestPackage(pkgName, pkgPath) {
-  return runInPackage({
```
adamweeks: @bzang this was failing and no matter what params I passed got it to work. This way works now, not sure if it just never got updated with other utils changes.

bzang: Oof, yea sorry about that one. There's a bit of goofiness with that when I moved the other scripts to be more synchronous.

packages/node_modules/@ciscospark/redux-module-conversation/src/actions.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+describe(`redux-module-conversation actions`, () => {
+  it(`has exported actions`, () => {
+    expect(actions.acknowledgeActivityOnServer).toBeDefined();
+    expect(actions.createConversation).toBeDefined();
+    expect(actions.deleteActivity).toBeDefined();
+    expect(actions.getConversation).toBeDefined();
+    expect(actions.loadMissingActivities).toBeDefined();
+    expect(actions.loadPreviousMessages).toBeDefined();
+  });
```
bzang: ooo, good call on this test.

packages/node_modules/@ciscospark/redux-module-conversation/src/actions.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+  it(`can acknowledge an activity`, () => {
+    const store = mockStore({});
+    const activity = {id: `activity-to-acknowledge`};
+    return store.dispatch(actions.acknowledgeActivityOnServer({toJS: jest.fn()}, activity, mockSpark))
```
bzang: Maybe create a constructor for the Map instead of using `{toJS: jest.fn()}`? Took me way too long to figure out what that was.

adamweeks: Haha, yea, good point. We should refactor our actions to not have to rely on immutable objects or not at some point.

PR #153, opened 2017-04-27
feat(widget-space): Space and WMM events update

bzang:
Update docs to add new eventing, and also add ability to pass eventNames as props.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
+++ b/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
@@ -167,7 +167,7 @@ export function hangupCall(call) {
   return (dispatch) => {
     // Don't update call states after hangup
     call.off();
-    call.hangup().then(() => {
+    return call.hangup().then(() => {
       dispatch(callDisconnected());
     });
   };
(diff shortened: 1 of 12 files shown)


packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -167,7 +167,7 @@ export function hangupCall(call) {
   return (dispatch) => {
     // Don't update call states after hangup
     call.off();
-    call.hangup().then(() => {
+    return call.hangup().then(() => {
```
adamweeks: Why is this in a docs update PR?

packages/node_modules/@ciscospark/widget-meet/src/constants.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
adamweeks: Wrong PR?

adamweeks:
Did we want to do the `.activity` here instead of `initialActivity` or will that be later? Probably should decide on it before the final release of the docs

bzang:
That requires a more extensive code change than we probably want at this point. Was gonna wait until the implementation actually happens before updating the docs to make sure that the docs always represent the actual code.

PR #156, opened 2017-04-28
Convert widget demos into separate packages

adamweeks:


Jenkinsfile
```diff
@@ -146,7 +147,9 @@ ansiColor('xterm') {
 
             archive 'packages/node_modules/@ciscospark/widget-message-meet/dist/**/*'
             archive 'packages/node_modules/@ciscospark/widget-space/dist/**/*'
-            archive 'dist/**/*'
+            archive 'packages/node_modules/@ciscospark/widget-message-meet-demo/dist/**/*'
```
adamweeks: @bzang I'm 99% confident that the raw dist folder was for the demo, but if not, let me know!

bzang: Yup, we were only grabbing that for the demo.

jd-carroll:
@adamweeks @bzang How are developers supposed to run `npm start`?

adamweeks:
@jd-carroll right now you would run `npm run start:package` with the demo or package that you want to run (so if you want to run the widget space demo it would be `npm run start:package widget-space-demo`)

PR #159, opened 2017-05-02
feat: message send failure and retry ability

adamweeks:
Now, with added unit tests!


packages/node_modules/@ciscospark/react-component-activity-item-base/src/styles.css
```diff
@@ -44,10 +52,23 @@
   flex: 0 0 auto;
 }
 
-.pending {
```
adamweeks: Had to change the pending opacity to each element because we didn't want the error message to opaque even though it was a sub element of `.pending`

packages/node_modules/@ciscospark/react-component-activity-item-base/src/index.js
```diff
@@ -15,12 +15,14 @@ export default function ActivityItemBase(props) {
     children,
     id,
     isAdditional,
+    isError,
```
bzang: nitpick: `isError` sounds a bit off ... `hasError` sounds more grammatically natural, but is inconsistent.

packages/node_modules/@ciscospark/react-component-activity-item-base/src/styles.css
bzang: Makes sense

packages/node_modules/@ciscospark/react-component-activity-list/src/index.js
```diff
@@ -29,7 +29,7 @@ export default function ActivityList(props) {
       );
     }
     case ITEM_TYPE_ACTIVITY: {
-      const {activity, avatarUrl, isPending, isAdditional, isFlagged, isSelf} = visibleActivity;
+      const {activity, avatarUrl, isError, isPending, isAdditional, isFlagged, isSelf} = visibleActivity;
```
bzang: What do you think about using destructuring on line 17 to set the values there instead of doing it for every case?

packages/node_modules/@ciscospark/react-component-utils/src/activity.js
```diff
@@ -5,7 +5,7 @@ import {isImage} from './files';
  * Constructs a default activity
  *
  * @export
- * @param {any} conversation
+ * @param {object} conversation
```
bzang: I don't think conversation is guaranteed to be a object right? I think this is set up to also take conversation as just a conversation id.

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
```diff
@@ -184,9 +218,36 @@ export function submitActivity(conversation, activity, user, spark) {
  * @returns {function}
  */
 export function setUserTyping(isTyping, conversation, spark) {
+  return (dispatch) =>
+    spark.conversation.updateTypingStatus(conversation.toJS(), {typing: isTyping})
```
bzang: does this need a `return`?

packages/node_modules/@ciscospark/redux-module-activity/src/actions.test.js
```diff
@@ -0,0 +1,214 @@
...
+
+const DATE_TO_USE = new Date(`2016`);
+const _Date = Date;
+global.Date = jest.fn(() => DATE_TO_USE);
+global.Date.UTC = _Date.UTC;
+global.Date.parse = _Date.parse;
+global.Date.now = _Date.now;
+
+window.URL.createObjectURL = jest.fn();
+window.URL.revokeObjectURL = jest.fn();
```
bzang: Should these mocks and global variables be set in the top level jest configs so that they're available everywhere?

packages/node_modules/@ciscospark/redux-module-activity/src/reducer.js
```diff
@@ -53,7 +61,10 @@ export default function reducer(state = initialState, action) {
 
   case RESET_ACTIVITY: {
     const inFlightActivities = state.get(`inFlightActivities`);
-    return initialState.setIn([`inFlightActivities`], inFlightActivities);
+    const activityFailures = state.get(`activityFailures`);
+    return initialState
+      .setIn([`inFlightActivities`], inFlightActivities)
+      .setIn([`activityFailures`], activityFailures);
```
bzang: Are these supposed to be `.set()` calls?

packages/node_modules/@ciscospark/react-component-activity-item-base/src/index.js
adamweeks: Yea, I think `hasError` makes sense here because we are saying the item "is pending and has an error". I'll change that.

packages/node_modules/@ciscospark/react-component-activity-list/src/index.js
adamweeks: This is the the only case that needs these values and it is from the `visibleActivity` object that is only available after the `.map`.

packages/node_modules/@ciscospark/react-component-utils/src/activity.js
adamweeks: We're not using it that way, but we are doing `id: conversation.id || conversation,` below. We need the full conversation object for retries now, so I'll remove the ability to just pass an ID.

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
adamweeks: it is via implicit return 😄

packages/node_modules/@ciscospark/redux-module-activity/src/reducer.js
adamweeks: img

PR #160, opened 2017-05-02
Fix read receipts

adamweeks:


packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
```diff
@@ -7,7 +7,7 @@ export const getMostRecentReadReceipts = createSelector(
   [getActivities, getParticipants],
   (activities, participants) => {
     const activity = activities.last();
-    return participants.filter((participant) =>
+    return participants.toJS().filter((participant) =>
```
adamweeks: I could've done ```participant.get(`roomProperties`)``` within the filter, but then we would still need to do the `toJS` method since the read-receipts component is expecting a basic javascript object.

packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
```diff
@@ -126,13 +126,10 @@ function receiveMercuryActivity(state, action) {
...
     participants = state.get(`participants`).map((participant) => {
-      if (participant.id === actorId) {
-        return Object.assign({}, participant, {
-          roomProperties: Object.assign({}, participant.roomProperties, {
-            lastSeenActivityUUID: activity.object.id,
-            lastSeenActivityDate: activity.published
-          })
-        });
+      if (participant.get(`id`) === actorId) {
+        return participant
```
adamweeks: Immutable reducing the amount of code!

PR #172, opened 2017-05-08
feat(react-redux-spark-metrics): Add Metrics module

bzang:
Not ready for primetime yet


adamweeks:
@bzang let me know when this is ready, looks like it is passing CI now

bzang:
@adamweeks, i think we're good to go with the first version of this

packages/node_modules/@ciscospark/react-redux-spark-metrics/src/reducer.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default function reducer(state = initialState, action) {
+  switch (action.type) {
+  case STORE_METRIC:
+    return state.setIn([`items`, action.payload.name], fromJS(action.payload.data));
+  case DELETE_METRIC:
+    return state.deleteIn([`items`, action.payload.name]);
+  case ADD_TO_QUEUE: {
+    const newQueue = state.get(`queue`).push(fromJS(action.payload.data));
+    return state.set(`queue`, newQueue);
```
adamweeks: I'm surprised there's no `pushIn` that we could use.

packages/node_modules/@ciscospark/widget-space-demo/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@ciscospark/widget-space",
+  "name": "@ciscospark/widget-space-demo",
```
adamweeks: Good catch.

PR #174, opened 2017-05-09
test(journeys): add data-api tests

adamweeks:


test/journeys/server/data-api/space.html
```diff
@@ -0,0 +1,28 @@
...
+    id="ciscospark-widget">
+
+</div>
+<script>
+  window.loadBundle = function (bundleFile) {
+    var script = document.createElement('script');
+    script.src = bundleFile;
+    document.head.appendChild(script);
+    script.onload = function () {
+      console.info(`bundle loaded`);
```
bzang: Back ticks may be an issue in non chrome browsers. Should probably just keep everything in JS to single ticks.

test/journeys/specs/messageMeet/widget-message-meet/dataApi.js
```diff
@@ -0,0 +1,137 @@
...
+      `spark:memberships_read`,
+      `spark:memberships_write`,
+      `spark:messages_read`,
+      `spark:messages_write`,
+      `spark:teams_read`,
+      `spark:teams_write`,
+      `spark:team_memberships_read`,
+      `spark:team_memberships_write`,
+      `spark:kms`
+    ].join(` `);
```
bzang: Is this ready to be moved to spark:all yet?

test/journeys/server/data-api/space.html
adamweeks: Yea, good call.

test/journeys/specs/messageMeet/widget-message-meet/dataApi.js
adamweeks: We may still need `webexsquare.get_conversation` for some things, but I think `spark:all` should handle it. I'll try it out.

PR #176, opened 2017-05-10
fix(widget-recents): fix split error, membership:created trigger, unread for new convos

bzang:


adamweeks:
@bzang 
>/home/ubuntu/react-ciscospark/packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
  12:16  error  Function 'SpaceItem' has a complexity of 8  complexity

PR #177, opened 2017-05-10
fix(react-container-read-receipts): fix vertical stacking

adamweeks:


packages/node_modules/@ciscospark/react-container-read-receipts/src/index.js
```diff
@@ -45,7 +49,7 @@ export class ReadReceipts extends Component {
       });
 
     return (
-      <div>
+      <div className={classNames(`ciscospark-read-receipts`, styles.readReceipts)}>
```
bzang: This fixes vertical stacking, but if there are too many users in the room then all the avatars still show. Do we wanna truncate? Looks like clients show first 10 and then a circle with the number of people left

adamweeks: gif
Feature Creep! 😛

adamweeks: I created a ticket for Meeta to prioritize.

PR #179, opened 2017-05-11
feat(widget-space): Space events

bzang:
Add new events model to spaces widget


packages/node_modules/@ciscospark/widget-meet/src/events.js
```diff
@@ -0,0 +1,21 @@
+import {base64} from '@ciscospark/common';
+
+export const eventNames = {
+  CALLS_CREATED: `calls:created`,
+  CALLS_CONNECTED: `calls:connected`,
+  CALLS_DISCONNECTED: `calls:disconnected`
+};
+
+export function constructCallEventData(call, actor, spaceId) {
```
adamweeks: jsdoc? 😛

packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -37,18 +37,20 @@ import {
 import styles from './styles.css';
 import {getMeetWidgetProps} from './selector';
 
-import events from './constants';
+import {
+  eventNames,
```
adamweeks: Could we do `eventNames as defaultEventNames` so we don't get it confused with `this.eventNames`?

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -57,7 +56,11 @@ import styles from './styles.css';
 import messages from './messages';
 
 import {getMessageWidgetProps} from './selector';
-import events from './constants';
+import {
+  eventNames,
```
adamweeks: `as defaultEventNames` here too?

PR #184, opened 2017-05-11
feat(widget-recents): Incoming calls UI

bzang:
UI and event handling for incoming calls.


packages/node_modules/@ciscospark/react-component-incoming-call/src/index.js
```diff
@@ -44,7 +44,7 @@ export default function IncomingCall({
         {displayName}
       </div>
       <div className={classNames(styles.incomingCallLabel, `incoming-call-label`)}>
-        {formatMessage(messages.incomingCallMessage)}
+        {incomingCallMessage}
```
adamweeks: Nice! I like moving the intl stuff out

packages/node_modules/@ciscospark/widget-recents/src/reducer.js
```diff
@@ -3,20 +3,33 @@ import mercury from '@ciscospark/redux-module-mercury';
...
-import {UPDATE_STATUS} from './actions';
+import {
+  UPDATE_STATUS,
+  STORE_INCOMING_CALL,
+  DELETE_INCOMING_CALL,
+  UPDATE_WEBRTC_SUPPORT
+} from './actions';
 
 export const initialState = fromJS({
+  incomingCall: null,
```
adamweeks: Should we make it `incomingCalls`? What happens if we have multiple calls coming in at the same time? I know the UI will only display one, but if once we answer or decline the first one, we should display the second next.

bzang: Yea, this doesn't take that into consideration. I think the better way to do this would be to refactor `module-media` to support all of this. There are a couple of other UI elements that weren't designed, like what happens when you load the recents widget and there are active calls that you can join?

adamweeks: Gotcha.

PR #185, opened 2017-05-11
fix(widget-meet): Fix webrtc detection in safari

bzang:
Some style fixes to make it look prettier


packages/node_modules/@ciscospark/widget-message-meet/src/styles.css
```diff
@@ -110,7 +110,6 @@
 .activity-component-wrapper > div {
   display: flex;
   width: inherit;
-  height: inherit;
```
bzang: Fixes safari UI issue

adamweeks: `#thankssafari`

PR #186, opened 2017-05-12
chore: install specific version of recompose so demo works

adamweeks:
we can revert this after this is released:
https://github.com/callemall/material-ui/pull/6847


adamweeks:
https://github.com/acdlite/recompose/issues/380

PR #187, opened 2017-05-12
Recents Updates

adamweeks:
Bugfixes and optimizations from SSDK-720 and others.

Also, welcome `prop-types`!

### DIFF
--- a/packages/node_modules/@ciscospark/react-component-spinner/src/index.js
+++ b/packages/node_modules/@ciscospark/react-component-spinner/src/index.js
@@ -6,11 +6,7 @@ import classNames from 'classnames';
 import styles from './styles.css';
 
 export default function spinner({bright}) {
-  const classes = [`ciscospark-spinner`, styles.spinner];
-  if (bright) {
-    classes.push(styles.bright);
-  }
-  return <div className={classNames(classes)} />;
+  return <div className={classNames(`ciscospark-spinner`, styles.spinner, {[styles.bright]: bright})} />;
 }
 
 spinner.propTypes = {
--- a/packages/node_modules/@ciscospark/react-component-spinner/src/index.js
+++ b/packages/node_modules/@ciscospark/react-component-spinner/src/index.js
@@ -1,8 +1,18 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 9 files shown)


packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
```diff
@@ -128,8 +129,15 @@ export function initialFetchSpaces(sparkInstance) {
...
-      conversationsLimit: 50
+      conversationsLimit: 10
     }))
+    .then(({spaces}) => {
+      const maxDate = _.last(spaces).lastReadableActivityDate;
+      return dispatch(fetchSpaces(sparkInstance, {
+        maxDate,
+        conversationsLimit: 50
+      }));
+    })
```
bzang: Not sure we can use _.last to determine max date. AFAIK spaces returned are not sorted by `lastReadableActivityDate`.

packages/node_modules/@ciscospark/react-component-spinner/src/index.js
```diff
@@ -1,8 +1,18 @@
...
 import styles from './styles.css';
 
-export default function spinner() {
-  return <div className={classNames(`ciscospark-spinner`, styles.spinner)} />;
+export default function spinner({bright}) {
+  const classes = [`ciscospark-spinner`, styles.spinner];
+  if (bright) {
+    classes.push(styles.bright);
+  }
+  return <div className={classNames(classes)} />;
```
bzang: TIL classnames can be flagged, so another way to do this check is:
`classNames('ciscospark-spinner', styles.spinner, {[styles.bright]: bright})`

packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
adamweeks: Oh great. Should I just skip the since portion and just load 50 after the first 10?

packages/node_modules/@ciscospark/react-component-spinner/src/index.js
adamweeks: Oh nifty! I'll change that to that format.

PR #204, opened 2017-05-14
fix(widget-message): Fix html filter

bzang:
This should fix our html cleaning issue and match closely with what the clients have in place.
We filter both on outgoing and anything we set dangerously in incoming.
Will need to create a config option to pass the filtered values in a configurable way.

### DIFF
--- a/packages/node_modules/@ciscospark/react-component-activity-text/src/index.test.js
+++ b/packages/node_modules/@ciscospark/react-component-activity-text/src/index.test.js
@@ -4,13 +4,20 @@ import renderer from 'react-test-renderer';
 import ActivityText from '.';
 
 describe(`ActivityText component`, () => {
-  const content = `Test Text`;
+  const displayName = `Test Text`;
+  const content = `<em>Test Text</em>`;
 
-  const component = renderer.create(
-    <ActivityText content={content} />
-  );
+  it(`renders clean text properly`, () => {
+    const component = renderer.create(
+      <ActivityText displayName={displayName} />
+    );
+    expect(component).toMatchSnapshot();
+  });
 
-  it(`renders properly`, () => {
+  it(`renders dangerous text properly`, () => {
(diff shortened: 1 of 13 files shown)


packages/node_modules/@ciscospark/react-component-activity-text/src/index.test.js
```diff
@@ -4,13 +4,20 @@ import renderer from 'react-test-renderer';
...
-  );
+  it(`renders clean text properly`, () => {
+    const component = renderer.create(
+      <ActivityText displayName={displayName} />
+    );
+    expect(component).toMatchSnapshot();
+  });
 
-  it(`renders properly`, () => {
+  it(`renders dangerous text properly`, () => {
```
adamweeks: Should we add one more test with some html that should be cleaned?

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
```diff
@@ -300,30 +300,45 @@ function revokeObjectURL(file) {
...
+    const messageObject = {
+      displayName: message
+    };
+    const markedString = marked(message) || ``;
+    if (markedString !== message) {
+      // Remove possibly added <p></p> tags and newline
+      const cleanedString = markedString.replace(/(?:^<p>)|(?:<\/p>$)|(?:<\/p>\n$)/g, ``).trim();
+
+      if (cleanedString !== messageString) {
+        messageObject.content = filterSync(() => {}, { // eslint-disable-line no-empty-function
```
adamweeks: Like you said in chat, we should get this into a reusable object at some point.

packages/node_modules/@ciscospark/react-component-activity-text/src/index.test.js
bzang: yup, good call

PR #205, opened 2017-05-15
fix: add static version of webrtc-adapter

adamweeks:


adamweeks:
https://github.com/webrtc/adapter/issues/536

PR #211, opened 2017-05-17
Error display component

adamweeks:
This is just the first implementation of the error display. Will be expanding upon this PR with a few more for different error situations.


packages/node_modules/@ciscospark/widget-space/src/container.js
```diff
@@ -193,7 +195,30 @@ export class SpaceWidget extends Component {
...
+      let errorSubtitle;
+      switch (error.name) {
+      case `Bad Space Id`:
+        errorSubtitle = formatMessage(messages.badSpaceId);
+        break;
+      case `NotFound`:
+        errorSubtitle = formatMessage(messages.errorNotFound);
+        break;
+      default:
+        errorSubtitle = formatMessage(messages.unknownError);
```
adamweeks: These were the only errors I could get and since there is no official documentation on the errors, we're stuck with this. Would be good to put some sort of metrics or logging on any errors that don't match.

bzang: Gotcha, yea, we're working on getting Sentry in place somewhere so that we can start capturing exceptions

packages/node_modules/@ciscospark/widget-space/src/actions.js
```diff
@@ -1,4 +1,22 @@
+import {isUuid} from '@ciscospark/react-component-utils';
+
+export const FETCH_SPACE_ERROR = `widget-space/FETCH_SPACE_ERROR`;
```
bzang: Would it make sense to put all of this into a redux-module, that way the logic doesn't have to be repeated each top level widget?

packages/node_modules/@ciscospark/widget-space/src/actions.js
```diff
@@ -45,19 +59,54 @@ export function storeActivityTypes(activityTypes) {
...
+    let spaceIdUUID;
+    if (isUuid(spaceId)) {
+      spaceIdUUID = spaceId;
+    }
+    else {
+      let isEncodedUUID = false;
+      let base64uuid;
+      // Check for base 64 encoding
+      try {
+        base64uuid = atob(spaceId).split(`/`).pop();
```
bzang: May want to put this into the utils package since we use it a lot. Also the other direction as well

packages/node_modules/@ciscospark/widget-space/src/container.js
bzang: Also can this logic go into a selector?

packages/node_modules/@ciscospark/widget-space/src/actions.js
adamweeks: True. Will do.

packages/node_modules/@ciscospark/widget-space/src/container.js
adamweeks: Yep, will move it over.

packages/node_modules/@ciscospark/widget-space/src/actions.js
adamweeks: I'm not sure what the module would do. Each one of these errors will probably be a unique case. Unless you meant moving the fetch space details into its own module.

bzang: Oh, I think i misinterpreted this specific case. This is actually the specific error that gets stored when the fetching a space errors out.

I was imagining generic keyed store for any error, and actions that pushed errors into that store. It could then be imported into any widget, and you could just check flags on that store to see if there's an error or not. Displaying the error would still be up to the widget itself, but there would be a global state.

That would make it so that when we start composing the widgets, the top level widget could read errors stored by the lower level widgets and react to them.

PR #213, opened 2017-05-18
Automation call fixes

adamweeks:


test/journeys/specs/messageMeet/widget-message-meet/meet.js
```diff
@@ -56,9 +56,24 @@ describe(`Widget Message Meet`, () => {
         toPersonEmail: localToUserEmail,
         initialActivity: `message`
       };
-      window.openWidgetMessageMeet(options);
+      window.openSpaceWidget(options);
     }, spock.token.access_token, mccoy.email);
```
bzang: Shouldn't this still open wmm?

adamweeks: Yep, that's why it failed, lol.

PR #212, opened 2017-05-17
chore: Update to use spark-js-sdk 0.8.0

bzang:
Related to
fixes #189
fixes #190 
fixes #191 
fixes #192 
fixes #193
fixes #194 
fixes #195 
fixes #196 
fixes #197 
fixes #198 
fixes #199 
fixes #200 
fixes #201 
fixes #202 
fixes #203

PR #214, opened 2017-05-18
Space Error for no auth

adamweeks:
This also checks `error.message` to give a proper error title.


adamweeks:
This will fail until #212 merges.

adamweeks:
Ready for review.

packages/node_modules/@ciscospark/widget-space/src/selector.js
```diff
@@ -69,14 +56,52 @@ export const getActivityTypes = createSelector(
   }
 );
 
+const getWidgetError = createSelector(
+  [getWidget, getSpark, getFormatMessage],
+  (widget, spark, formatMessage) => { // eslint-disable-line complexity
+    const registerError = spark.getIn([`status`, `registerError`]);
+    const widgetError = widget.getIn([`status`, `isError`]);
```
bzang: Could be imagining that we talked about this, but didn't we say we wanted this to be `hasError`?

adamweeks: Yea, I don't think it was this code, but we definitely wanted to use `hasError`. I'll change that.

adamweeks:
Once the recents code gets tackled, I'll figure out how to refactor into reusable code bits, but leaving it like this for now.

PR #215, opened 2017-05-19
test(journeys): sauce stabilization and logging

adamweeks:


wdio.conf.js
```diff
@@ -4,6 +4,9 @@ let mochaTimeout = 30000;
 if (process.env.DEBUG_JOURNEYS) {
   mochaTimeout = 99999999;
 }
+if (process.env.SAUCE) {
+  mochaTimeout = 90000;
```
adamweeks: I think this might be the solution!

PR #235, opened 2017-05-23
fix(react-redux-spark): stop attempting to register on failure

adamweeks:
If you give a bad token, it will constantly try to register device.


adamweeks:
I don't see any. This basically fixes a bug introduced in #214

PR #238, opened 2017-05-23
Display error for no connection

adamweeks:
Was considering going with the injectErrors way that you did metrics. Thoughts on this store method compared to it?

### DIFF
--- a/packages/node_modules/@ciscospark/widget-space/src/actions.js
+++ b/packages/node_modules/@ciscospark/widget-space/src/actions.js
@@ -75,13 +75,11 @@ export function getSpaceDetails(sparkInstance, spaceId, intl) {
     if (!spaceIdUUID) {
       const displayTitle = formatMessage(messages.unableToLoad);
       const displaySubtitle = formatMessage(messages.badSpaceId);
-      const code = `Bad Space Id`;
       return dispatch(addError({
         id: `getSpace.badId`,
         displayTitle,
         displaySubtitle,
-        temporary: false,
-        code
+        temporary: false
       }));
     }
     dispatch(fetchingSpaceDetails());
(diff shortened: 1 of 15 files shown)


bzang:
I'm not actually 100% happy with how the metrics were done. The reason why I created the HOC was because I needed a way to process a queue on renders and to have different behavior when the sparkInstance wasn't available yet.

If errors is just a store with a few helper functions to push and pull errors out, then maybe using our basic redux-module pattern is enough. The actions exported should be able to handle everything. I'd only do the HOC if you need to depend on the sparkInstance.

adamweeks:
Ok, that is what I was thinking. The hardest part was figuring out how to get spark registration errors. I didn't really want to put it in the spark store actions, so I put it in widget space's setup to check for spark store errors and then create the error message to display.

packages/node_modules/@ciscospark/widget-space/src/actions.js
```diff
@@ -70,23 +65,47 @@ function fetchingSpaceDetails() {
...
     const spaceIdUUID = validateAndDecodeId(spaceId);
     if (!spaceIdUUID) {
-      const badSpaceId = `Bad Space Id`;
-      const error = new Error(badSpaceId);
-      error.name = badSpaceId;
-      return dispatch(fetchSpaceError(error));
+      const displayTitle = formatMessage(messages.unableToLoad);
+      const displaySubtitle = formatMessage(messages.badSpaceId);
+      const code = `Bad Space Id`;
+      return dispatch(addError({
```
adamweeks: We're adding the errors to the store as soon as they happen instead of doing it in the selector

packages/node_modules/@ciscospark/widget-space/src/container.js
```diff
@@ -249,6 +306,17 @@ export class SpaceWidget extends Component {
         </div>
       );
     }
+    else if (errors.get(`hasError`)) {
```
adamweeks: Adding this error display here as well because we may get an error before we have a spark instance or space details.

adamweeks:
@bzang this is ready for review now

PR #251, opened 2017-05-24
feat(react-container-read-receipts): using selector to create props

adamweeks:


packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
```diff
@@ -1,14 +1,36 @@
+/* eslint-disable max-params */
 import {createSelector} from 'reselect';
 
 const getActivities = (state) => state.conversation.get(`activities`);
 const getParticipants = (state) => state.conversation.get(`participants`);
+const getCurrentUser = (state) => state.user.get(`currentUser`);
```
bzang: It looks like currentUser is a normal object. If that's the case then I think this will cause the selector to return a new object every time and force the main selector to rerun every time. May want to just grab `user` and then grab `currentUser` from inside the selector.

adamweeks: I was monitoring the selector being fired and actually the opposite was true. It fired WAY more often without getting the current user.

bzang: Oh nice! There's more comparison logic in there than I thought.

adamweeks: Yea, it triggered every time the user store updated, with this, only when current user changed.

PR #257, opened 2017-05-24
chore(package): update jest to version 20.0.4

bzang:
Closes #168


adamweeks:
Looks like you'll need to update snapshots since this is now immutable aware.

bzang:
@adamweeks donzo!

PR #258, opened 2017-05-26
tests(tap): fix tap tests for wmm

bzang:


test/journeys/specs/messageMeet/widget-message-meet/basic.js
```diff
@@ -27,7 +27,7 @@ describe(`Widget Message Meet`, () => {
 
   before(`load browsers`, () => {
     browser
-      .url(`/message-meet.html`)
+      .url(`/production.html`)
```
adamweeks: This will cause our wmm tests to not be run during builds

bzang: Oops, I got over exuberant. Thanks for catching that.

PR #266, opened 2017-06-01
chore: update jenkinsfile.tap

momaney:


Jenkinsfile.tap
```diff
@@ -12,12 +12,6 @@ pipeline {
   }
 
   stages {
-    stage('Checkout') {
-      steps {
-        git 'https://github.com/ciscospark/react-ciscospark.git'
-      }
-    }
-
     stage('Clean') {
```
adamweeks: We probably don't need this stage either

PR #274, opened 2017-06-07
testing: Redux testing

bzang:
Added missing action tests to all redux modules.


bzang:
Gonna clean up the immutable stores first before tackling the reducer tests.

PR #276, opened 2017-06-07
Read receipt grouping

adamweeks:


adamweeks:
@bzang let's make a new ticket for tooltips. We need those on the regular avatars in read receipts as well.

PR #281, opened 2017-06-09
Avatar fetching

adamweeks:
This turned out a bit bigger than expected, so you may want to go commit by commit.

Basically, all individual avatar handling is now in the avatar module. You can call "fetch avatar" over and over for a user and it will only process if we haven't seen that user before. 

There's also some cleanup in spaces widget to only get the avatars we need for display instead of **EVERYONE**

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
+++ b/packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
@@ -50,17 +50,9 @@ export function initialFetchSpaces(sparkInstance) {
 
 export function fetchSpaceAvatar(sparkInstance, space) {
   return (dispatch) => {
-    const currentUserId = sparkInstance.internal.device.userId;
     dispatch(fetchingAvatar(space.id));
     if (space.type === `direct`) {
-      // Support for direct messages with users who no longer exist
-      if (space.participants.length > 1) {
-        const otherUserId = space.participants.find((p) => p.id !== currentUserId).id;
-        return sparkInstance.people.get(otherUserId)
-          .then((person) => {
-            dispatch(storeAvatar(space.id, {url: person.avatar}));
-          });
-      }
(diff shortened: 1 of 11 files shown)


packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
```diff
@@ -50,17 +50,9 @@ export function initialFetchSpaces(sparkInstance) {
...
     if (space.type === `direct`) {
-      // Support for direct messages with users who no longer exist
-      if (space.participants.length > 1) {
-        const otherUserId = space.participants.find((p) => p.id !== currentUserId).id;
-        return sparkInstance.people.get(otherUserId)
-          .then((person) => {
-            dispatch(storeAvatar(space.id, {url: person.avatar}));
-          });
-      }
+      return Promise.reject(new Error(`Direct spaces should use avatar service`));
```
bzang: We should probably use the avatar module to retrieve and store all avatars, not just the user avatars.

adamweeks: I'm going to create another tech debt ticket for that. I was looking into doing that with this PR but I was 100% on it.

packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -74,16 +82,17 @@ function constructGroup(space, avatars) {
 }
 
 const getRecentSpaces = createSelector(
-  [getSpaces, getUsers],
-  (spaces, users) => {
+  [getSpaces, getUsers, getAvatars],
+  (spaces, users, avatars) => {
     const currentUser = users.get(`currentUser`);
-    const avatars = spaces.get(`avatars`);
+    const spaceAvatars = spaces.get(`avatars`);
```
adamweeks: For now.

PR #285, opened 2017-06-12
chore(tooling): Add csp to webpack dev server

bzang:
So that we can catch any csp issues if something changes.


adamweeks:
We should do a test build environment for automation as well. Doesn't have to be in this PR if you don't want to.

bzang:
Yea, can you create a ticket for that and outline the unique requirements of the test env setup.

adamweeks:
Done. https://voxeolabs.atlassian.net/browse/SSDK-945

PR #304, opened 2017-06-17
refactor(widget-space): convert to object for spaceDetails

bzang:
This is a patch for a Proxy issue

### DIFF
--- a/packages/node_modules/@ciscospark/widget-space/src/reducer.js
+++ b/packages/node_modules/@ciscospark/widget-space/src/reducer.js
@@ -46,7 +46,7 @@ export function reducer(state = initialState, action) {
 
   case STORE_SPACE_DETAILS:
     return state.setIn([`status`, `isFetchingSpaceDetails`], false)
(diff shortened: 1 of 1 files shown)


packages/node_modules/@ciscospark/widget-space/src/reducer.js
```diff
@@ -46,7 +46,7 @@ export function reducer(state = initialState, action) {
 
   case STORE_SPACE_DETAILS:
     return state.setIn([`status`, `isFetchingSpaceDetails`], false)
-      .set(`spaceDetails`, fromJS(action.payload.details));
+      .set(`spaceDetails`, fromJS(Object.assign({}, action.payload.details)));
```
adamweeks: Would a merge work better here?

bzang: Unfortunately not if the `action.payload.details` is a Proxy object. Also we don't necessarily want to merge it in if it's a new space we're dealing with.

PR #305, opened 2017-06-19
Event data - actorName and tests expanded

adamweeks:
Doing some detailed tests on the events that arrive and their data structure.


adamweeks:
https://voxeolabs.atlassian.net/browse/SSDK-949

PR #309, opened 2017-06-20
Space avatars moved to avatar module

adamweeks:
https://voxeolabs.atlassian.net/browse/SSDK-938


packages/node_modules/@ciscospark/redux-module-avatar/src/actions.js
```diff
@@ -0,0 +1,91 @@
...
+  };
+}
+
+/**
+ * Fetches an avatar for a given user id
+ * @param {string} userId
+ * @param {object} spark
+ * @returns {Thunk}
+ */
+export function fetchAvatarForUserId(userId, spark) {
```
bzang: Do we still need this if we have `fetchSpaceAvatar`? Seems like all the logic is repeated. We could have something like a single `fetchAvatar({userIds: [], space}, sparkInstance)`

PR #310, opened 2017-06-20
test(widget-space): Add file transfer tests tests

ricalver:
includes pdf, txt, doc, docx, ppt, html, json, zip, gif, jpg, png, and mp3

jira ticket: SSDK-588


test/journeys/specs/oneOnOne/messaging.js
```diff
@@ -7,6 +7,10 @@ import '@ciscospark/internal-plugin-conversation';
 import {switchToMessage} from '../../lib/menu';
 import {clearEventLog, getEventLog} from '../../lib/events';
 import {constructHydraId} from '../../lib/hydra';
+const request = require(`superagent`);
```
adamweeks: Let's use the `import` method here instead of the `require` to keep things consistent.

test/journeys/specs/oneOnOne/messaging.js
```diff
@@ -114,13 +122,94 @@ describe(`Widget Space: One on One`, () => {
 
     it(`sends and deletes message`);
 
-    it(`sends message with pdf attachment`);
+    describe(`File Transfer Tests`, () => {
+      /* eslint no-sync: "off" */
+      const uploadDir = path.join(__dirname, `assets`);
+      const downloadDir = path.join(os.homedir(), `Downloads`);
+      const inputFileElem = `.ciscospark-file-input`;
+      const downloadFileButton = `(//div[@title="Download this file"]/parent::button)[last()]`;
```
adamweeks: Do we not have an aria label for the download button? If not, we need to make a ticket to do so.

test/journeys/specs/oneOnOne/messaging.js
```diff
@@ -114,13 +122,94 @@ describe(`Widget Space: One on One`, () => {
...
+        const fileSize = fs.statSync(filePath).size;
+        const downloadedFile = path.join(downloadDir, fileName);
+        const fileTitle = `//div[text()="${fileName}"]`;
+        if (fs.existsSync(downloadedFile)) {
+          fs.unlinkSync(downloadedFile);
+        }
+        browserLocal.chooseFile(inputFileElem, filePath);
+        browserLocal.click(shareButtonElem);
+        browserRemote.waitForExist(fileTitle, 30000);
+        browserRemote.moveToObject(downloadFileButton);
```
adamweeks: How are we moving to the object then waiting for it to be visible? Does it move to where it is supposed to be?

test/journeys/specs/oneOnOne/messaging.js
```diff
@@ -114,13 +122,94 @@ describe(`Widget Space: One on One`, () => {
...
+        const events = getEventLog(browserRemote);
+        const newMessage = events.find((event) => event.eventName === `messages:created`);
+        const fileUrl = newMessage.detail.data.files[0].url;
+        let downloadedFileSize;
+        request.get(fileUrl)
+            .set(`Authorization`, `Bearer ${mccoy.token.access_token}`)
+            .end((error, response) => {
+              downloadedFileSize = response.header[`content-length`];
+            });
+        browser.pause(8000);
```
adamweeks: Take a look at the `wait-for-promise.js` file to see if that will work instead of pausing the browser. Just in case 8 seconds isn't long enough.

test/journeys/specs/oneOnOne/messaging.js
ricalver: I checked, and it does not have an aria label. Should we get that fixed before I finish addressing these comments?

test/journeys/specs/oneOnOne/messaging.js
ricalver: The button has to have a mouseover event for it to appear, and become clickable.  MoveToObject accomplishes the mouse over, but click can occur faster than the button becomes visible, resulting in a false negative.  So, I added to waitForVisible to prevent that.

test/journeys/specs/oneOnOne/messaging.js
ricalver: The idea with pause was to have a timeout setting at the download level.  I can proceed with wait-for-promise, but I worry about the possibility of a socket hanging, causing the remaining tests to fail on the global 30 second timeout, resulting in false negatives.   Let me know if we still want to use wait-for-promise with that considered.

test/journeys/specs/oneOnOne/messaging.js
adamweeks: No, I've created a ticket and will make sure to update the tests as well once it is done. https://voxeolabs.atlassian.net/browse/SSDK-1051

test/journeys/specs/oneOnOne/messaging.js
adamweeks: That's not very nice of them. Could you add a code comment about this?

test/journeys/specs/oneOnOne/messaging.js
adamweeks: This will work for now, as long as we don't need more timeout for sauce tests. 

I believe you can modify the timeout for a specific test suite too, but I haven't done it yet.

adamweeks:
Once we get that require changed over and the code comment about the mouse move, we should be good to go!

wdio.conf.js
```diff
@@ -23,7 +23,8 @@ exports.config = {
       `./test/journeys/specs/tap/**/*.js`
     ],
     oneOnOne: [
-      `./test/journeys/specs/oneOnOne/**/*.js`
+      // `./test/journeys/specs/oneOnOne/**/*.js`
```
adamweeks: Did you mean to change this? None of the other suites will run if so.

ricalver: lol, crap.  That's just so I can test faster.  I'll change it back.

test/journeys/specs/oneOnOne/messaging.js
adamweeks: Now that I think about it, let's `moveToObject` the parent container that causes the button to appear. This will be what the user will be doing as well. The user wouldn't know to move to the hidden button and I'm worried that we are taking advantage of a webdriver bug.

PR #328, opened 2017-06-27
fix(widget-space): Button style collision

bzang:
Ended up only fixing the specific issue. Main issue is that we're passing classes down into components which breaks css modules.

### DIFF
--- a/packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/index.js
+++ b/packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/index.js
@@ -26,10 +26,12 @@ export default function MessageMeetActivityMenu({
     }
   ];
   return (
-    <ActivityMenu
-      buttons={buttons}
-      onExit={onExit}
-    />
+    <div className={styles.activityMenuContainer}>
+      <ActivityMenu
+        buttons={buttons}
+        onExit={onExit}
+      />
+    </div>
   );
 }
 
--- a/packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/styles.css
+++ b/packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/styles.css
@@ -1,7 +1,11 @@
-.message-button {
+.activity-menu-container {
+  height: 100%;
+}
+
+.activity-menu-container .message-button {
   background-color: #07c1e4;
 }
 
-.meet-button {
+.activity-menu-container .meet-button {
   background-color: #30d557;
 }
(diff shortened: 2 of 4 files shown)


packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/index.js
```diff
@@ -26,10 +26,12 @@ export default function MessageMeetActivityMenu({
     }
   ];
   return (
-    <ActivityMenu
-      buttons={buttons}
-      onExit={onExit}
-    />
+    <div className={styles.activityMenuContainer}>
```
adamweeks: Do we need a user configurable style name here?

packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/styles.css
```diff
@@ -1,7 +1,11 @@
...
+.activity-menu-container {
+  height: 100%;
+}
+
+.activity-menu-container .message-button {
   background-color: #07c1e4;
 }
 
-.meet-button {
+.activity-menu-container .meet-button {
```
adamweeks: Good call

scripts/webpack/webpack.base.babel.js
```diff
@@ -18,7 +18,12 @@ export default (options) => {
     ]),
     new ExtractTextPlugin({filename: `[name].css`, disable: false, allChunks: true}),
     // Adds use strict to prevent catch global namespace issues outside of chunks.
-    new webpack.BannerPlugin(`react-ciscospark v${packageJson.version}`)
+    new webpack.BannerPlugin(`react-ciscospark v${packageJson.version}`),
+    new webpack.LoaderOptionsPlugin({
```
adamweeks: Should we put a note that this is just a temporary change until the plugins support context config?

bzang: I couldn't find any reference for when they'd add it. It seems to just be the accepted "fix".

packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/index.js
bzang: There's a parent container that already has a user configurable name... this is just so we have some context in the component to grab onto with our scoped styles.

scripts/webpack/webpack.base.babel.js
adamweeks: Oh lord. https://webpack.js.org/plugins/loader-options-plugin/ says
>This plugin will be removed in the future as it only exists for migration.

packages/node_modules/@ciscospark/widget-message-meet/src/components/activity-menu/index.js
adamweeks: Ok, I'm sure we'll need to go through at some point and make sure everything is user configurable, but that is WAY down the priority list.

scripts/webpack/webpack.base.babel.js
bzang: Oh snap! Well guess we probably shouldn't keep that in there then.

PR #331, opened 2017-06-28
refactor: Babel build

bzang:
Add building to ES5 and Common JS. Looks big but most files are just package.json adjustments to support different build targets.


package.json
```diff
@@ -3,15 +3,15 @@
...
-    "clean": "npm run clean:snapshots && npm run clean:dist",
+    "build": "cross-env NODE_ENV=${NODE_ENV:-production} ./scripts/build/index.js",
+    "build:all": "npm run build all",
+    "build:package": "npm run build dist",
+    "build:packagejson": "npm run build package.json",
+    "build:widgets": "npm run build widgets",
+    "clean": "npm run clean:snapshots && npm run clean:dist && npm run clean:transpile",
     "clean:snapshots": "rimraf \"packages/**/__snapshots__/\"",
     "clean:dist": "rimraf \"packages/**/dist/\"",
+    "clean:transpile": "rimraf \"packages/**/es/\" && rimraf \"packages/**/cjs/\"",
```
ianwremmel: should this just be part of clean:dist?

package.json
```diff
@@ -3,15 +3,15 @@
...
-    "build:packagejson": "cross-env NODE_ENV=${NODE_ENV:-production} babel-node --presets=stage-1,es2015 scripts/update-package-json.js",
-    "build:widgets": "cross-env NODE_ENV=${NODE_ENV:-production} babel-node --presets=stage-1,es2015 scripts/build-widgets.js",
-    "clean": "npm run clean:snapshots && npm run clean:dist",
+    "build": "cross-env NODE_ENV=${NODE_ENV:-production} ./scripts/build/index.js",
+    "build:all": "npm run build all",
+    "build:package": "npm run build dist",
+    "build:packagejson": "npm run build package.json",
+    "build:widgets": "npm run build widgets",
+    "clean": "npm run clean:snapshots && npm run clean:dist && npm run clean:transpile",
     "clean:snapshots": "rimraf \"packages/**/__snapshots__/\"",
```
ianwremmel: unrelated: why do snapshots need to be cleaned?

package.json
```diff
@@ -3,15 +3,15 @@
...
-    "build": "npm run build:packages && npm run build:bundle",
-    "build:packages": "cross-env NODE_ENV=${NODE_ENV:-production} babel-node --presets=stage-1,es2015 scripts/build-packages.js",
-    "build:package": "cross-env NODE_ENV=${NODE_ENV:-production} babel-node --presets=stage-1,es2015 scripts/build-package.js",
-    "build:bundle": "cross-env NODE_ENV=${NODE_ENV:-production} babel-node --presets=stage-1,es2015 scripts/build-bundle.js",
-    "build:packagejson": "cross-env NODE_ENV=${NODE_ENV:-production} babel-node --presets=stage-1,es2015 scripts/update-package-json.js",
-    "build:widgets": "cross-env NODE_ENV=${NODE_ENV:-production} babel-node --presets=stage-1,es2015 scripts/build-widgets.js",
-    "clean": "npm run clean:snapshots && npm run clean:dist",
+    "build": "cross-env NODE_ENV=${NODE_ENV:-production} ./scripts/build/index.js",
+    "build:all": "npm run build all",
+    "build:package": "npm run build dist",
```
ianwremmel: shouldn't these be `npm run build -- dist`?

scripts/build/commands/all.js
```diff
@@ -0,0 +1,28 @@
...
+} = require(`../../utils/build`);
+const {getAllPackagePaths} = require(`../../utils/package`);
+
+
+module.exports = {
+  command: `all`,
+  desc: `Build all packages`,
+  builder: {},
+  handler: () =>
+    getAllPackagePaths().map((pkg) => {
```
ianwremmel: This is going to run each package in parallel. when lerna did stuff like that, it tried to melt my cpu; should these be more sequential?

scripts/utils/build.js
```diff
@@ -0,0 +1,142 @@
...
+  // copy if it's a css file
+  else if (ext === `.css`) {
+    return outputFileSync(outputPath, content, {encoding: `utf8`});
+  }
+  return false;
+}
+
+
+function babelBuild(folderPath, destination, babelOptions = {}, firstFolder = true) {
+  const stats = fs.statSync(folderPath);
```
ianwremmel: pretty sure this'll throw if `folderPath` doesn't exist. does that matter?

scripts/utils/deps.js
```diff
@@ -1,17 +1,20 @@
...
-export default function updatePackageJson() {
+/**
+ * Updates all package.json files with depedencies
+ * @returns {undefined}
+ */
+function updatePackageJson() {
   const topPkgJson = JSON.parse(readFileSync(`./package.json`, `utf8`));
   const packages = getAllPackages();
   const pkgPaths = getAllPackagePaths();
+  console.log(packages);
```
ianwremmel: log statement?

ianwremmel: the `debug` package is really handle for optional logging of build steps

package.json
bzang: Turns out `--` really isn't necessary anymore... not sure when that happened.

package.json
bzang: This is a piece of legacy script when we were still tweaking how we were building snapshots and the jest script couldn't support `-u` for snapshot updates.

package.json
ianwremmel: interesting. I think I've found `--` isn't needed in basic use cases, but if you start throwing switches in for your own scripts (or possibly multiple params), `npm run` gets really confused.

scripts/build/commands/all.js
bzang: AFAI can tell, this actually it doesn't do it in parallel. I think it's because the rest of our build scripts are synchronous. The component packages all get babelified in about 10 sec. The webpack components are the ones that take a really long time.

scripts/utils/deps.js
bzang: Hmmm good call. I'll add that in on the next PR with the css modules fix.

scripts/build/index.js
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Build Targets
+ * Bundle : Single JS file with everything
+ * CommonJS : Compiled source with all transforms and polyfills
+ */
+
+require(`yargs`) // eslint-disable-line no-unused-expressions
+  .usage(`Usage: $0 <target> [args]`)
+  .commandDir(`commands`)
+  .demandCommand(1, `Please let us know what you'd like to build.`)
```
adamweeks: So friendly!

PR #338, opened 2017-06-30
chore: Babel css modules

bzang:
Allow babel to build for css modules.

Caveat, currently if we're building es5 w/ import/export the css modules are not being transpiled correctly correctly.  This PR fixes the issue: https://github.com/wbyoung/babel-plugin-transform-postcss/pull/28

Also needed to update to node v7 to support. May need @momaney or @aarcher11 to update the docker container with node 7.


postcss.config.js
```diff
@@ -1,8 +1,16 @@
...
+module.exports = (ctx) => ({
   plugins: {
+    'postcss-modules': {
+      getJSON: ctx.extractModules || (() => {})
+    },
     'postcss-cssnext': {
       browsers: [`last 2 versions`, `IE > 10`]
     },
+    cssnano: {
+      preset: `default`,
```
adamweeks: What is this doing? Just better minification?

bzang: cssnano does some interesting magic that consolidates styles and tries to look for repeated styles to consolidate.

PR #341, opened 2017-07-04
fix(widget-space): Allow user to call when direct spaceID is provided

bzang:
Not a great patch, but looks like our `widget-meet` structure is a bit off. This will fix the bug, but we probably need to clean up and refactor the meet container at some point.


packages/node_modules/@ciscospark/widget-space/src/actions.js
```diff
@@ -5,6 +5,7 @@ import messages from './messages';
 
 
 export const FETCH_SPACE_ERROR = `widget-space/FETCH_SPACE_ERROR`;
+export const STORE_TO_PERSON = `widget-space/STORE_TO_PERSON`;
```
ianwremmel: maybe at some future point we should reconsider "to person". i initially read this "point store to a person", not "store the 'to person'"

bzang: Haha... yea, we need to actually refactor the `redux-module-users`. One that happens we can actually just point to the stored user in whatever widgets we build.

PR #346, opened 2017-07-06
fix(widget-recents): Avatar refresh bug and slightly improve overall avatar performance

bzang:


packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -153,6 +135,59 @@ export class RecentsWidget extends Component {
   }
 
   @autobind
+  fetchAllAvatar(nextProps) {
```
adamweeks: Nitpicky: Should this be `fetchAllAvatars`?

adamweeks:
Are the scripts supposed to be in this PR?

bzang:
Was a bit of an in place refactor because I got tired of typing out ` npm run start:package widget-`

PR #348, opened 2017-07-06
styles(widget-space): Updated space row to pointer cursor on hover

bzang:


bzang:
@ianwremmel is there a DOM element that does what a `button` does but feels a tad better than `button`? I think it may be good to attach a `role="button"`.

PR #354, opened 2017-07-14
feat(react-component-activity-text): use linkify to create links

adamweeks:


packages/node_modules/@ciscospark/react-component-activity-text/src/index.js
```diff
@@ -38,19 +43,22 @@ export default function ActivityText(props) {
...
     }, [], content);
-    const htmlContent = {__html: cleaned};
+    const linkifiedContent = linkifyHtml(cleaned, linkifyOptions);
+    const htmlContent = {__html: linkifiedContent};
     return (
       <div
         className={classNames(`ciscospark-activity-text`, styles.activityText)}
         dangerouslySetInnerHTML={htmlContent}
       />
     );
```
bzang: Probably don't need this return since the next block returns the same jsx. Just need to set the HTML content based on if the HTML content is available or not

packages/node_modules/@ciscospark/react-component-activity-text/src/index.js
```diff
@@ -38,19 +43,22 @@ export default function ActivityText(props) {
       h5: [],
       h6: []
     }, [], content);
-    const htmlContent = {__html: cleaned};
+    const linkifiedContent = linkifyHtml(cleaned, linkifyOptions);
```
bzang: How does this work if the other clients return a `a` tag? Can linkify handle that or does it double wrap the link?

packages/node_modules/@ciscospark/widget-space/src/index.html
```diff
@@ -26,7 +26,7 @@
       data-space-id="<%= process.env.SPACE_ID %>"
     />
   </div>
-</div>
+</div> -->
```
bzang: Commented out?

packages/node_modules/@ciscospark/react-component-activity-text/src/index.js
adamweeks: Oh yea, good catch. Changed the second block after the first one.

packages/node_modules/@ciscospark/widget-space/src/index.html
adamweeks: img

packages/node_modules/@ciscospark/react-component-activity-text/src/index.js
adamweeks: The `linkifyHtml` is aware of the html within and shouldn't change it. Just tested and it works, but I will add a test suite to make sure.

adamweeks:
@bzang ready for round 2

PR #373, opened 2017-07-25
chore(tooling): Add try...catch retry to Tap Tests

bzang:
This should force a retry on any type of error and also attach failure logs to the body

### DIFF
--- a/Jenkinsfile.tap
+++ b/Jenkinsfile.tap
@@ -67,7 +67,7 @@ pipeline {
 
   post {
     failure {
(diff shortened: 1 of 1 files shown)


Jenkinsfile.tap
```diff
@@ -1,5 +1,27 @@
...
 
+def tapTest = { ->
+  sh '''#!/bin/bash -ex
+  source ~/.nvm/nvm.sh
+  nvm use v7
+  CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true npm run test:tap
+  '''
+}
+
+// We only retry once. If it fails again we
```
adamweeks: `If it fails again we` we what!?! The suspense is killing me!

PR #379, opened 2017-08-01
Fix css global

adamweeks:
Also threw in a refactor to remove react-intl requirement in a component since we already had the message defined.

### DIFF
--- a/packages/node_modules/@ciscospark/react-component-list-separator/src/styles.css
+++ b/packages/node_modules/@ciscospark/react-component-list-separator/src/styles.css
@@ -1,14 +1,6 @@
-:root {
+.separator {
   --color-primary: #049fd9;
   --color-gray: #aaa;
-}
-
-.separator.informative {
-  color: var(--color-primary);
-  text-transform: uppercase;
-}
-
-.separator {
   display: flex;
   padding-top: 17px;
   overflow: hidden;
@@ -18,6 +10,12 @@
   white-space: nowrap;
   align-items: center;
 }
+
+.separator.informative {
+  color: var(--color-primary);
+  text-transform: uppercase;
+}
+
 
 .separator-text {
   position: relative;
--- a/packages/node_modules/@ciscospark/react-component-new-messages-separator/src/index.js
+++ b/packages/node_modules/@ciscospark/react-component-new-messages-separator/src/index.js
@@ -1,22 +1,18 @@
 import React from 'react';
-import {FormattedMessage} from 'react-intl';
+import PropTypes from 'prop-types';
 
 import ListSeparator from '@ciscospark/react-component-list-separator';
 
-function NewMessageSeparator() {
-  const newMessages = ( // eslint-disable-line no-extra-parens
-    <FormattedMessage
-      defaultMessage={`NEW MESSAGES`}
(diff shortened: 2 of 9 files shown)


packages/node_modules/@ciscospark/react-component-list-separator/src/styles.css
```diff
@@ -1,14 +1,6 @@
-:root {
+.separator {
   --color-primary: #049fd9;
```
bzang: Does this just scope the variables to the .separator class?

packages/node_modules/@ciscospark/react-component-new-messages-separator/src/index.js
```diff
@@ -1,22 +1,18 @@
...
-function NewMessageSeparator() {
-  const newMessages = ( // eslint-disable-line no-extra-parens
-    <FormattedMessage
-      defaultMessage={`NEW MESSAGES`}
-      description={`Indicator to display that new, unread messages follow`}
-      id={`newMessages`}
-    />
-  );
-
+function NewMessageSeparator({message}) {
```
bzang: Nice! I like that we're moving the translations up.

packages/node_modules/@ciscospark/react-component-new-messages-separator/src/styles.css
```diff
@@ -1,8 +1,5 @@
-:root {
-  --color-primary: #049fd9;
-}
-
 .new-messages {
+  --color-primary: #049fd9;
   color: var(--color-primary);
```
bzang: This seems a big weird that we're declaring a variable and then just setting the color to it. What is the point?

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -613,6 +613,7 @@ export class MessageWidget extends Component {
               {
                 user.get(`currentUser`) &&
                 <ActivityList
+                  newMessagesMessage={formatMessage(messages.newMessagesMessage)}
```
bzang: Lol, we need a different word for 'Message'

packages/node_modules/@ciscospark/react-component-list-separator/src/styles.css
adamweeks: Yep, and everything below it.

packages/node_modules/@ciscospark/react-component-new-messages-separator/src/styles.css
adamweeks: I don't think there was reasoning behind the variables honestly. Couldn't tell you my motivation especially since it isn't reused. I could just remove them.

packages/node_modules/@ciscospark/widget-message/src/container.js
adamweeks: I don't understand your message. Maybe you can message me your message and I'll display it in a message window. 😛

packages/node_modules/@ciscospark/react-component-new-messages-separator/src/styles.css
adamweeks: LOL. This file isn't even being used. Removing.

PR #397, opened 2017-08-04
fix(widget-space): Fix initialActivity setting with data-api

bzang:


bzang:
oops fat fingered that one.

bzang:
Let's try that again @adamweeks

adamweeks:
There we go!

PR #400, opened 2017-08-07
Feature flags

adamweeks:


packages/node_modules/@ciscospark/redux-module-features/src/index.js
```diff
@@ -0,0 +1,45 @@
...
+export default function reducer(state = initialState, action) {
+  switch (action.type) {
+  case ADD_FEATURE:
+    return state.mergeIn([`items`], action.payload.feature);
+  default:
+    return state;
+  }
+}
+
+export function addFeature(feature) {
```
bzang: Can we rename this to `storeFeature`? It took me a while to realize that this just saves the features we get from the server.

packages/node_modules/@ciscospark/redux-module-features/src/index.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Gets the value of the requested feature toggle.
+ * @param {string} keyType <developer|user|entitlement>
+ * @param {string} key
+ * @param {object} spark
+ * @returns {Promise}
+ */
+export function getFeature(keyType, key, spark) {
```
bzang: Should we have a "getAllFeatures" method to grab all the features for a user? Or is the preferred way to fetch individual ones?

packages/node_modules/@ciscospark/redux-module-features/src/index.js
adamweeks: Sure thing.

packages/node_modules/@ciscospark/redux-module-features/src/index.js
adamweeks: @ianwremmel suggested to get specific features explicitly.

ianwremmel: yea, the full set is unbounded; i think i have several hundred at this point. the web client does copy them all from the sdk and wire up change listeners, but that's mostly there to provide a UI. if you only need to read them, getting them one at a time should be safer.

PR #420, opened 2017-08-09
feat(widget-meet): display error for security requirement

adamweeks:


packages/node_modules/@ciscospark/widget-meet/src/messages.js
```diff
@@ -31,9 +31,13 @@ export default defineMessages({
     defaultMessage: `Cisco Spark cannot find your camera. Please check your browser settings, and that your camera is accessible, and try again.`
   },
   callErrorNotAllowed: {
-    id: `ciscospark.container.meet.message.deviceNotFound`,
+    id: `ciscospark.container.meet.message.errorNotAllowed`,
     defaultMessage: `Cisco Spark cannot access your camera. Please check your browser settings and try again.`
   },
+  callErrorSecurity: {
+    id: `ciscospark.container.meet.message.errorSecurity`,
+    defaultMessage: `Cisco Spark cannot access your camera. Video Calling is supported on secure (HTTPS) sites only.`
```
bzang: Is https a camera issue? I don't think we need "Cisco Spark cannot access your camera".

adamweeks: Good call.

PR #421, opened 2017-08-09
chore: add spark upgrade script

adamweeks:
We'll use this script as part of the js-sdk post deploy jenkins script.

### DIFF
--- a/Jenkinsfile.jssdk
+++ b/Jenkinsfile.jssdk
@@ -46,15 +46,20 @@ ansiColor('xterm') {
 
         stage('Branching') {
           echo "Creating Branch"
-          def version = sh script: 'npm view ciscospark version', returnStdout: true
-          version = version.trim()
+          def version
           sh '''#!/bin/bash -ex
+          source ~/.nvm/nvm.sh
+          nvm use v7
+          version=$(node ./scripts/utils/jssdkversion.js)
+          echo $version > .version
           git add package.json
           git commit -m "chore: update @ciscospark packages to ${version}"
           '''
+          version = readFile '.version'
           sshagent(['6c8a75fb-5e5f-4803-9b6d-1933a3111a34']) {
-            sh "git push upstream HEAD:js-sdk/${version}"
+            sh "echo git push upstream HEAD:js-sdk/${version}"
           }
+          archiveArtifacts artifacts: 'package.json'
         }
       }
     }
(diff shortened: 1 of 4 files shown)


adamweeks:
@ianwremmel yep, thoughts on making it autodeploy? Or should it just upgrade, install, run all tests and mark the build as good?

ianwremmel:
I would have Jenkins open a pr but still keep it a manual merge, at least for version one.

Jenkinsfile.jssdk
```diff
@@ -0,0 +1,44 @@
...
+
+        stage('Branching') {
+          echo "Creating Branch"
+          sh '''#!/bin/bash -ex
+          version=`npm view ciscospark version`
+          git add package.json
+          git commit -m "chore: update @ciscospark packages to ${version}"
+          '''
+          sshagent(['6c8a75fb-5e5f-4803-9b6d-1933a3111a34']) {
+            sh "git push upstream HEAD:js-sdk/${version}"
```
adamweeks: Not familiar with this, but is the `version` param available within sshagent?

Jenkinsfile.jssdk
```diff
@@ -0,0 +1,44 @@
...
+          source ~/.nvm/nvm.sh
+          nvm use v7
+          npm run upgradespark
+          '''
+        }
+
+        stage('Branching') {
+          echo "Creating Branch"
+          sh '''#!/bin/bash -ex
+          version=`npm view ciscospark version`
```
ianwremmel: use $() instead of backticks. same effect, but nestable, so it's a good habit to get into.

Jenkinsfile.jssdk
ianwremmel: nope. the version variable is only available in the original sh block. 

you could store it with a line like

```groovy
def version = sh script: 'npm view ciscospark version', returnStdout: true
version = version.trim()
```

Jenkinsfile.jssdk
```diff
@@ -0,0 +1,44 @@
...
+            nvm install 7
+            nvm use v7
+            npm install
+            rm -f $HOME/.npmrc
+            '''
+          }
+        }
+
+        stage('Upgrade') {
+          echo "Upgrading @ciscospark packages"
```
bzang: Do we need to pull in the newest master branch first before doing this?

Jenkinsfile.jssdk
adamweeks: Added.

Jenkinsfile.jssdk
adamweeks: Good call.

adamweeks: Added.

PR #428, opened 2017-08-14
chore(package): update file-loader to version 1.0.0

adamweeks:
Closes #376


bzang:
Can we change the target to hotfix-20170926?

PR #445, opened 2017-08-30
feat: Add notification onclick events

bzang:


packages/node_modules/@ciscospark/react-container-notifications/src/reducer.js
```diff
@@ -9,7 +9,7 @@ import {
 export const initialState = fromJS({
   items: {},
   settings: {
-    isSupported: false,
+    isSupported: null,
```
adamweeks: Good call.

packages/node_modules/@ciscospark/widget-recents/events.md
```diff
@@ -195,3 +205,31 @@ The user has either left (`actorId` is yourself) or has been removed (`actorId`
   }
 }
 ```
+
```
adamweeks: Nice! Thanks for updating the docs to go along with the changes!

PR #447, opened 2017-09-01
feat(widget-recents): add toPersonEmail to event data

adamweeks:
I've created a build from this PR. Should we go through the full release process or point it at `develop`?


adamweeks:
@bzang good call, will do

PR #451, opened 2017-09-05
fix: Audio is muted when video is muted

bzang:
Currently merging into `develop` but may need to hotfix into `master`


adamweeks:
@bzang is this the right branch? I don't see where you split the streams

bzang:
Redoing this PR

PR #452, opened 2017-09-06
fix: Video muting, mutes audio

bzang:


packages/node_modules/@ciscospark/react-component-video/src/index.js
```diff
@@ -13,19 +13,29 @@ export default function Video({src, srcObject, audioMuted}) {
     }
   }
 
+  if (videoMuted) {
```
adamweeks: We probably don't want this feature any more since we are splitting tracks. Otherwise, we'll have two audio tags on the page.

adamweeks:
Oh, want to throw some basic jest tests at the `Audio` component?

packages/node_modules/@ciscospark/react-component-video/src/index.js
bzang: Good call

PR #456, opened 2017-09-11
fix(widget-recents): Remove no-enter keypresses for selecting space

bzang:
Fix for SSDK-1264

### DIFF
--- a/packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
+++ b/packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
@@ -25,7 +25,11 @@ export default function SpaceItem({
   type
 }) {
 
-  function handleClick() {
+  function handleClick(e) {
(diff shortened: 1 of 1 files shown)


packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
```diff
@@ -25,7 +25,11 @@ export default function SpaceItem({
   type
 }) {
 
-  function handleClick() {
+  function handleClick(e) {
```
adamweeks: Is this supposed to be in the "handle click" function? What happens if you actually click?

bzang: Yea, i cheated a bit here... probably should have a separate method for handle keypress

adamweeks: Yea gotcha. I usually do too

PR #459, opened 2017-09-12
fix: Hiding one-on-one rooms user has left

bzang:
Fixes SSDK-1265


packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -1,5 +1,6 @@
 import {createSelector} from 'reselect';
 import moment from 'moment';
+import {OrderedMap} from 'immutable';
```
adamweeks: Looks good. When I get back, I want to go over my selector strategy. Basically, I remove all of the immutable items out and only pass real javascript objects to the container.

PR #468, opened 2017-09-19
feat(react-hoc-conversation-mercury): refactor mercury conversation events

adamweeks:
✅ Automation tests for widget-space pass

### DIFF
--- a/packages/node_modules/@ciscospark/react-hoc-conversation-mercury/package.json
+++ b/packages/node_modules/@ciscospark/react-hoc-conversation-mercury/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "@ciscospark/react-hoc-conversation-mercury",
+  "description": "Cisco Spark React Conversation Mercury Higher Order Component",
+  "main": "./cjs/index.js",
+  "src": "./src/index.js",
+  "module": "./es/index.js",
+  "keywords": [],
+  "contributors": [
+    "Adam Weeks <adweeks@cisco.com>",
+    "Bernie Zang <nzang@cisco.com>"
+  ],
+  "license": "MIT",
+  "repository": "https://github.com/ciscospark/react-ciscospark",
+  "files": [
+    "dist",
+    "src"
+  ]
+}
--- a/packages/node_modules/@ciscospark/widget-message/src/container.js
+++ b/packages/node_modules/@ciscospark/widget-message/src/container.js
@@ -19,7 +19,7 @@ import {
   loadPreviousMessages,
   receiveMercuryActivity,
   receiveMercuryComment,
-  updateMercuryState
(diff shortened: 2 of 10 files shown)


packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -294,20 +294,18 @@ export class MessageWidget extends Component {
...
-    props.updateMercuryState({isListeningToActivity: true});
+    props.updateWidgetState({isListeningToActivity: true});
     sparkInstance.internal.mercury.on(`event:conversation.activity`, (event) => {
       const activity = event.data.activity;
       const isSelf = activity.actor.id === props.user.get(`currentUser`).id;
       // Ignore activity from other conversations
       if (activity.target && activity.target.id === space.id) {
         if (activity.object.objectType === `activity`) {
-          props.receiveMercuryActivity(activity);
           if (isSelf && activity.verb === `acknowledge`) {
```
bzang: Can we combine the 2 sequential if statements into a single one?

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -722,14 +720,12 @@ export default connect(
...
     retryFailedActivity,
     setScrollPosition,
     setTyping,
     showScrollToBottomButton,
     updateHasNewMessage,
-    updateMercuryState
+    updateWidgetState
   }, dispatch)
-)(MessageWidget);
+)(ConnectedMessageWidget);
```
bzang: Is this the right order of applying `connect()`? I can't tell if this is right or the order should be reversed like:
``` js
const MessageWidget = connect(getMessageWidgetProps, blah blah)(MessageWidget);
export default wrapConversationMercury(MessageWidget);

File: packages/node_modules/@ciscospark/react-hoc-conversation-mercury/package.json [hunk 3]
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@ciscospark/react-hoc-conversation-mercury",
+  "version": "0.1.0",
```
bzang: Let's remove versions from the packages `package.json` files. They don't mean anything, but seem to have caused some confusion for devs.

packages/node_modules/@ciscospark/widget-message/src/container.js
adamweeks: I would think we would need to connect the wrapped component so it gets all the redux stores it needs.

packages/node_modules/@ciscospark/widget-message/src/container.js
adamweeks: Oh yea, will do.

PR #469, opened 2017-09-20
fix(widget-recents): Loading performance

bzang:
Performance hotfixes so that we're loading recents in a staggered way into the DOM. A bit long, but mostly straightforward. Will try to make this a hotfix.


bzang:
This PR should clear up a lot more once we have the previous hotfix merged into Master.

packages/node_modules/@ciscospark/redux-module-avatar/src/actions.js
```diff
@@ -29,15 +29,10 @@ function addAvatarBegin(id) {
...
  */
 export function fetchAvatarsForUsers(userIds, spark) {
-  const promises = [];
-  return (dispatch) => {
-    userIds.forEach((userId) => {
-      promises.push(dispatch(fetchAvatar({userId}, spark)));
-    });
-    return Promise.all(promises);
-  };
+  return (dispatch) => Promise.all(userIds.map((userId) => dispatch(fetchAvatar({userId}, spark))));
```
adamweeks: I guess you could do it this way if you want it to be all fancy 😛

packages/node_modules/@ciscospark/redux-module-avatar/src/actions.js
```diff
@@ -57,40 +52,55 @@ export function fetchAvatar({space, userId}, spark) {
...
+  };
 }
 
-function fetchSpaceAvatar(space, spark, userIdForAvatar) {
-  if (space.type === `direct`) {
-    if (!userIdForAvatar) {
-      return Promise.reject(new Error(`Direct spaces require a user id to display`));
+function fetchSpaceAvatar(space, sparkInstance, userIdForAvatar) {
+  return (dispatch) => {
+    if (space.type === `direct` && space.participants.length === 2) {
```
adamweeks: Did you find a condition where a direct space had something other than 2 participants?

packages/node_modules/@ciscospark/redux-module-avatar/src/actions.js
```diff
@@ -57,40 +52,55 @@ export function fetchAvatar({space, userId}, spark) {
...
+    }
+    return Promise.resolve(dispatch(addAvatar(space.id)));
+  };
+}
+
+/**
+ * A bit of a hack to get smaller Avatars from Public API
+ */
+
+function convertToSmallAvatar(avatarUrl) {
```
adamweeks: Awesome, I never got around to putting this in!

packages/node_modules/@ciscospark/redux-module-teams/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+    isFetching: false,
+    hasFetched: false
+  }
+});
+
+
+export default function reducer(state = initialState, action) {
+  switch (action.type) {
+  case STORE_TEAMS:
+    return state.setIn([`status`, `hasFetched`], true)
```
adamweeks: May want to set `isFetching` to false here as well.

packages/node_modules/@ciscospark/redux-module-teams/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+    return state;
+  }
+}
+
+
+export function fetchTeams(sparkInstance) {
+  return (dispatch) => {
+    dispatch(updateStatus({isFetching: true}));
+    return sparkInstance.internal.team.list()
+      .then((teams) => {
```
adamweeks: Might want to `.catch` here too and set an error state.

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -140,51 +148,49 @@ export class RecentsWidget extends Component {
...
-            // Get the user ID of the participant that isn't current user
-            const currentUserId = user.get(`currentUser`).id;
-            const otherUser = s.get(`participants`).find((p) => p.get(`id`) !== currentUserId);
-            if (otherUser) {
-              const otherUserId = otherUser.get(`id`);
-              this.fetchAvatar({userId: otherUserId}, nextProps);
-            }
-          }
+      spacesList.forEach((s) => {
+        if (s.type === `direct` && s.participants.length > 1 && s.toPersonId) {
```
adamweeks: Are the spaces in `spacesList` an immutable object? Would you need to `get` those props?

packages/node_modules/@ciscospark/redux-module-avatar/src/actions.js
bzang: Yea, there were direct spaces that only had 1 participant, the logged in user. When that happened it would throw.

packages/node_modules/@ciscospark/redux-module-avatar/src/actions.js
bzang: 👍

packages/node_modules/@ciscospark/redux-module-teams/src/index.js
bzang: ooo yea, good catch.

packages/node_modules/@ciscospark/redux-module-teams/src/index.js
bzang: Yup yup.

packages/node_modules/@ciscospark/widget-recents/src/container.js
bzang: So I'll probably end up changing this in the future, but spacesList itself is immutable, but it's an ordered list of plain objects, so this works right now. Should probably go one way or the other (thinking just a plain object)

packages/node_modules/@ciscospark/redux-module-avatar/src/actions.js
adamweeks: Ah yea, that's right, when a user gets deleted

packages/node_modules/@ciscospark/widget-recents/src/container.js
adamweeks: That's why I've just been converting them all in the selector to regular js objects.

bzang: With this it's a bit different because the order matters, so OrderedMap is the best way to do this and plain js objects wouldn't maintain order while also allowing us to retrieve space data by ID

PR #470, opened 2017-09-20
fix(react-container-message-composer): Improve typing performance

bzang:
Added debounce and added checks for sending userTyping messages to server.

### DIFF
--- a/packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
+++ b/packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
@@ -21,12 +21,15 @@ import styles from './styles.css';
 export class MessageComposer extends Component {
   constructor(props) {
     super(props);
+    this.timerId = 0;
     this.handleTextChange = this.handleTextChange.bind(this);
     this.handleTextAreaBlur = this.handleTextAreaBlur.bind(this);
     this.handleTextAreaFocus = this.handleTextAreaFocus.bind(this);
     this.handleKeyDown = this.handleKeyDown.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
-    this.setUserTyping = debounce(this.setUserTyping.bind(this), 500, {leading: true, trailing: false});
+    this.setUserStartTyping = debounce(this.setUserStartTyping, 500, {leading: true, trailing: false});
+    this.setUserTyping = this.setUserTyping.bind(this);
+    this.setUserStopTyping = this.setUserStopTyping.bind(this);
   }
 
   shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
```diff
@@ -25,6 +26,7 @@ export class MessageComposer extends Component {
     this.handleTextAreaFocus = this.handleTextAreaFocus.bind(this);
     this.handleKeyDown = this.handleKeyDown.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
+    this.setUserTyping = debounce(this.setUserTyping.bind(this), 500, {leading: true, trailing: false});
```
adamweeks: How does debounce handle when we pass it different props? Would a `false` immediately send if there was a `true` waiting?

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
```diff
@@ -218,8 +218,10 @@ export function submitActivity(conversation, activity, user, spark) {
  * @returns {function}
  */
 export function setUserTyping(isTyping, conversation, spark) {
-  return (dispatch) =>
-    spark.internal.conversation.updateTypingStatus(conversation.toJS(), {typing: isTyping})
+  return (dispatch, getState) => {
+    const activity = getState().activity;
+    if (activity.getIn([`status`, `isTyping`]) !== isTyping) {
+      return spark.internal.conversation.updateTypingStatus(conversation.toJS(), {typing: isTyping})
```
adamweeks: We have to keep sending this because the server will auto-timeout and send a stop typing event if we don't.

bzang: Do we know what the timeout is?

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
bzang: Props don't matter, it only checks the function that is called. I think this is fine given that there's only a 500ms debounce.

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
adamweeks: I can't find it documented anywhere.

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
adamweeks: Ok, cool. Didn't want a true to be sent after a false if false was the last one to come in.

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
bzang: Ok, just got confirmation from the service owners that there is no auto-timeout. Clients handle all of this on their own.

bzang:
@adamweeks , ready for another round.

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
```diff
@@ -218,8 +218,10 @@ export function submitActivity(conversation, activity, user, spark) {
  * @returns {function}
  */
 export function setUserTyping(isTyping, conversation, spark) {
-  return (dispatch) =>
-    spark.internal.conversation.updateTypingStatus(conversation.toJS(), {typing: isTyping})
+  return (dispatch, getState) => {
+    const activity = getState().activity;
+    if (activity.getIn([`status`, `isTyping`]) !== isTyping) {
```
adamweeks: Since we have this check now in the container, do we want to get rid of the state check in the action and keep `getState` out of it? Up to you.

bzang: The check here is a bit different than the check in the container. The one in the container doesn't let you set isTyping to true multiple times. This prevents setting to false multiple times as well.

PR #477, opened 2017-09-22
chore(package): update jest to version 21.1.0

adamweeks:
Closes #450

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-presence/src/reducer.test.js
+++ b/packages/node_modules/@ciscospark/redux-module-presence/src/reducer.test.js
@@ -37,16 +37,16 @@ describe(`spark reducer`, () => {
         }
       }
     })).toMatchSnapshot();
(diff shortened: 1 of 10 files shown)


packages/node_modules/@ciscospark/redux-module-presence/src/reducer.test.js
```diff
@@ -37,16 +37,16 @@ describe(`spark reducer`, () => {
...
+  });
 
-    it(`should handle UPDATE_MODULE_STATUS`, () => {
-      expect(reducer(initialState, {
-        type: UPDATE_MODULE_STATUS,
-        payload: {
-          status: {
-            isSettingUserStatus: true
-          }
+  it(`should handle UPDATE_MODULE_STATUS`, () => {
```
adamweeks: somehow, nested `it`s worked until now

PR #479, opened 2017-09-22
chore(tooling): fix jenkinsfile.tap

moriahmaney:


Jenkinsfile.tap
```diff
@@ -1,63 +1,54 @@
...
-          throw error
         }
       }
     }
   }
+
+  post {
+    failure {
+      echo "build failed"
+      //emailext attachLog: true, body: 'Spark-js-sdk--react-widget--tap build number $BUILD_NUMBER has failed. For detailed information, please refer to $BUILD_URL', subject: 'Spark-js-sdk--react-widget--tap Failure', to: 'spark-js-sdk-gating-pipeline-email.94d222a7@ciscospark.pagerduty.com'
```
bzang: Looks like pagerduty email is commented out? Also could you test the formatter we added in the previous version? It should at least provide us with text in pager duty that tells us why things failed.

moriahmaney: Good catch, missed this. Lex probably had it commented out from when he was testing. As far as the formatter goes, I will have to ask about the priority of that as I'm still working with the dev ex team on things they need.

bzang: Check the previous version of the jenkinsfile. You should be able to copy it over and run it once to make sure there are no issues.

moriahmaney: Okay, will test it and then get a second CY approval.

aarcher11:
@moriahmaney please make the following changes as I verified these work

`
          retry(2){
            sh '''#!/bin/bash -ex
            set +x
            source ~/.nvm/nvm.sh
            nvm use v7
            set -x
            CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true npm run test:tap
            '''

post {
    failure {
      emailext body: '<p>$JOB_BASE_NAME build number $BUILD_NUMBER has failed. For detailed information, please refer to $BUILD_URL</p><pre>${BUILD_LOG_REGEX, regex="^Starting Tap Tests", linesBefore=0, linesAfter=10000, showTruncatedLines=false}</pre>', subject: 'Spark-js-sdk--react-widget--tap Failure', to: 'spark-js-sdk-gating-pipeline-email.94d222a7@ciscospark.pagerduty.com'
    }`

PR #486, opened 2017-09-26
Feat: Display At Mentions in Activity List

bzang:


.babelrc
```diff
@@ -12,7 +12,6 @@
     "lodash"
   ],
   "presets": [
-    "es2015",
```
adamweeks: This should stay, right?

packages/node_modules/@ciscospark/react-container-activity-list/src/formatters/at-mention.js
```diff
@@ -0,0 +1,73 @@
...
+    if (e.keyCode && e.keyCode === 13) {
+      return triggerEvent(e);
+    }
+    return false;
+  }
+
+  function triggerEvent(e) {
+    const target = e.target;
+    if (isMention(target)) {
+      onEvent(EVENT_NAME_MENTION_CLICKED, {
```
adamweeks: Should we check if onEvent exists first?

packages/node_modules/@ciscospark/react-container-activity-list/src/formatters/at-mention.js
```diff
@@ -0,0 +1,73 @@
...
+        onEvent={onEvent}
+      />
+    );
+
+    return Object.assign({}, activity, {component});
+  }
+  return activity;
+};
+
+function constructHydraId(type, id) {
```
adamweeks: We should eventually move this to a utility function, it is all over the codebase.

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.test.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable max-nested-callbacks */
+import {
+  formatActivity
+} from './selectors';
+
+describe(`formatActivity function`, () => {
```
adamweeks: img

packages/node_modules/@ciscospark/react-container-activity-list/src/formatters/at-mention.js
bzang: Agreed. I wanted to but didn't want to create another package in this PR for it.

.babelrc
bzang: Babel is pushing people to the `preset-env` setup and actually warning people when they install the `es2015` preset. That correctly transforms get pulled in based on our build targets.

packages/node_modules/@ciscospark/react-container-activity-list/src/formatters/at-mention.js
bzang: It's a bit obtuse but the `onEvent` check is in the `isMention()` check.

PR #492, opened 2017-09-28
fix(widget-message): Add check for buffer state listener

bzang:
Fix for error state when conversation is not available but buffer state has come back.


packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -329,16 +330,18 @@ export class MessageWidget extends Component {
    */
   @autobind
   listenToBufferState(conversationId, sparkInstance) {
```
adamweeks: `conversationId` probably isn't needed any more since we are using the conversation via props

PR #494, opened 2017-09-28
chore(package): update webpack to version 3.6.0

adamweeks:
Closes #434


adamweeks:
@bzang  Since this is out there, let's get the latest into hotfix

PR #525, opened 2017-10-04
fix(tooling): Fix babel preset es2015

bzang:
Only last commit matters


adamweeks:
Looks good, can you rebase on the hotfix now that it is updated with master to clean up commits?

PR #529, opened 2017-10-09
Feature: At mention submit

bzang:
Allow user to submit at mentions. Currently on allows people in the same room.


adamweeks:
@bzang could you throw some screen shots of the @ mention popup?

bzang:
Sure thing. @adamweeks 
initial
search
final

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
```diff
@@ -103,49 +115,128 @@ export class MessageComposer extends Component {
...
+      }
+    });
+
+    if (term) {
+      return participants.filter((p) => p.display.toLowerCase().includes(term.toLowerCase()));
+    }
+    return participants;
+  }
+
+  @autobind
```
adamweeks: I don't think we need `@autobind` here, do we?

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -100,10 +100,15 @@ export class MessageWidget extends Component {
...
   componentWillUpdate(nextProps) {
     const props = this.props;
+
     if (this.activityList
       && nextProps.conversation.get(`activities`).count() !== props.conversation.get(`activities`).count()) {
       this.scrollHeight = this.activityList.getScrollHeight();
     }
+
+    if (!nextProps.features.hasIn([`items`, `js-widgets-mentions`])) {
+      props.getFeature(`developer`, `js-widgets-mentions`, nextProps.sparkInstance);
```
adamweeks: `getFeature` isn't imported or in proptypes

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
bzang: Good catch

packages/node_modules/@ciscospark/widget-message/src/container.js
bzang: Oh wow, guess `getFeature` was getting passed by the container. I'll add that into this widget so we don't run into odd issues.

bzang:
@adamweeks, fixed. Ready for another round

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -736,6 +742,7 @@ export default connect(
     setTyping,
     showScrollToBottomButton,
     updateHasNewMessage,
-    updateWidgetState
+    updateWidgetState,
+    getFeature
```
adamweeks: Can you add it to the injected prop types as well?

bzang: Yup. Done

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -700,7 +700,8 @@ const injectedPropTypes = {
   updateHasNewMessage: PropTypes.func.isRequired,
   updateWidgetState: PropTypes.func.isRequired,
   space: PropTypes.object.isRequired,
-  features: PropTypes.object.isRequired
+  features: PropTypes.object.isRequired,
+  getFeature: PropTypes.object.isRequired
```
adamweeks: `getFeature` is a function 😭

PR #532, opened 2017-10-09
Roster search email

adamweeks:
This uses hydra for ALL email search, regardless of team.

### DIFF
--- a/packages/node_modules/@ciscospark/widget-roster/src/components/add-participant/index.js
+++ b/packages/node_modules/@ciscospark/widget-roster/src/components/add-participant/index.js
@@ -1,9 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import isEmail from 'validator/lib/isEmail';
 
 import classNames from 'classnames';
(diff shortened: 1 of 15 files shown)


packages/node_modules/@ciscospark/widget-roster/src/components/add-participant/index.js
```diff
@@ -27,6 +29,16 @@ export default function AddParticipant(props) {
     props.onAddPerson(person);
   }
 
+  function handleInvite() {
+    props.onAddPerson(props.searchTerm);
+  }
+
+  function handleKeyPressInvite(e) {
+    if (e.key === `Enter` || e.key === ` `) {
```
bzang: Does this add the person when spacebar is pressed? Doesn't spacebar usually page down?

adamweeks: It does when the person is selected and spacebar is pressed. According to a11y standards, buttons allow enter or space to select.

PR #533, opened 2017-10-10
Group calling for spaces

adamweeks:
- [x] No group icon in meet home
- [x] Call events changed redux-module-media actions.js
- [x] "no other participants" screen/state for group calls


packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -239,19 +260,25 @@ function constructCallState(call) {
 }
```
bzang: Should we remove these remote media objects? I think at least some of these don’t do anything anymore

packages/node_modules/@ciscospark/widget-meet/src/selector.js
```diff
@@ -44,18 +45,43 @@ export const getMeetWidgetProps = createSelector(
       }
     }
 
+    const hasCheckedWebRTCSupport = media.getIn([`webRTC`, `hasCheckedSupport`]);
+    const isWebRTCSupported = media.getIn([`webRTC`, `isSupported`]);
```
bzang: Nice!

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -295,10 +295,11 @@ export class MessageWidget extends Component {
     const {props} = this;
     const handleEvent = this.handleEvent;
     const toUser = space.toPerson;
+    const currentUserId = props.user.get(`currentUser`).id;
```
bzang: I like the pattern you used for the webrtc status check. We may want to move this stuff into the selector as well since we use it everywhere. Just have to be careful and make sure we’re memoizing correctly

adamweeks: Yea, I think that should be part of the render optimizations. IMO, we shouldn't be calling any immutable `get` methods in render.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -152,14 +152,18 @@ export function listenForIncomingCalls(sparkInstance, locusUrl) {
...
+
           // Once we are connected, our connected call state will handle disconnects
-          incomingCall.once(`connected`, () => incomingCall.off(`disconnected`, removeIncoming));
-          return bindEvents(dispatch, incomingCall);
+          incomingCall.on(`membership:change`, () => {
+            if (incomingCall.me.state === `connected`) {
+              incomingCall.off(`inactive`, removeIncoming);
+            }
+          });
+          return dispatch(storeCall(incomingCall));
```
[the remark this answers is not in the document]
adamweeks: I had a tough time with this one but maybe we should: 
```js
dispatch(storeCall(incomingCall))
  .then(() => incomingCall.acknowledge())
  .then(() => {
    // listen for connect events
  });
```

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
adamweeks: I don't see anything that isn't used.

PR #536, opened 2017-10-15
Feature: Add active call roster

bzang:
Added call roster after some refactoring. Doesn't look pretty right now, but it works. The files changed and commit history will clear up once the group calling PR makes it in.


adamweeks:
I rebased my group calling branch on master and all of the commits changed, probably why this one shows way more changes than it actually has. Want to try to rebase instead?

bzang:
@adamweeks There we go, much better!

packages/node_modules/@ciscospark/widget-roster/src/selector.js
```diff
@@ -1,111 +1,228 @@
...
+          }
+          else {
+            notInMeeting.push(participant);
+          }
+        }
+      });
+
+      if (inMeeting.length > 0) {
+        people.push({
+          label: formatMessage(messages.inMeeting),
```
adamweeks: Still feel like `formatMessage` is gross in a selector. Is there a way we could format the message in the container before passing it on to the people list? This works for now, but maybe in our performance improvements we could figure out a way.

bzang: Yea... it is gross. Haven't found a great way to do it yet for things like this. I'm hoping that once we do some normalizing of the reducers and stores that a better pattern will emerge.

PR #537, opened 2017-10-16
refactor(widget-message-meet): Remove widget

bzang:
Feels good to remove 2500 lines of code from the first time we tried to do this.


adamweeks:
Hooray! Let's check with @aarcher11 to make sure the CDN builds get these removed too.

PR #543, opened 2017-10-19
fix: Add check for immutable object is defined

bzang:
Added a bunch of `if()` checks before we perform a `toJS` to make sure the data exist.


packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
```diff
@@ -130,18 +130,20 @@ export class MessageComposer extends Component {
...
-    conversation.get(`participants`).toJS().forEach((p) => {
-      if (user.get(`currentUser`).id !== p.id) {
-        participants.push({
-          display: p.displayName,
-          id: p.id,
-          avatarUrl: avatar.getIn([`items`, p.id])
-        });
-      }
-    });
+    if (conversation.get(`participants`)) {
```
adamweeks: Might be better to do `conversation.has(` instead?

packages/node_modules/@ciscospark/react-container-notifications/src/selectors.js
```diff
@@ -17,7 +17,7 @@ const getUnsentNotifications = createSelector(
           return Object.assign({}, notification, {
             username: notificationActivity.actor.displayName,
             message: notificationActivity.object.displayName,
-            avatar: avatars[notificationActivity.actor.id]
+            avatar: avatars.get(notificationActivity.actor.id)
```
adamweeks: Does this make it return an immutable object for `avatar` instead of a js object?

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -167,7 +167,10 @@ export class RecentsWidget extends Component {
             nextProps.fetchAvatar({userId: s.toPersonId}, sparkInstance);
           }
           else {
-            nextProps.fetchAvatar({space: spaces.getIn([`items`, s.id]).toJS()}, sparkInstance);
+            const space = spaces.getIn([`items`, s.id]);
+            if (space) {
```
adamweeks: Could do a `hasIn` here

packages/node_modules/@ciscospark/widget-roster/src/selector.js
```diff
@@ -18,7 +18,7 @@ const getSearchResults = createSelector(
     const searchTerm = widgetRoster.get(`searchTerm`);
     if (searchTerm && search.hasIn([`searchResults`, searchTerm])) {
       const searchResults = search.getIn([`searchResults`, searchTerm]).toJS();
-      if (searchResults.results) {
+      if (searchResults && searchResults.results) {
```
adamweeks: Might be better to do ```search.hasIn([`searchResults`, searchTerm, `results`])``` now that I think about it.

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
bzang: Actually I think that passes as true every time as long as we've retrieved the conversation. `.has()` just checks if the key exists, so we should probably do `conversation.get(`participants`).first()` or `conversation.get(`participants`).count()`

packages/node_modules/@ciscospark/react-container-notifications/src/selectors.js
bzang: Not sure what was going through my brain when I was doing this one. Gonna revert it because this definitely doesn't make sense.

bzang:
Ready for another round @adamweeks

PR #544, opened 2017-10-20
Group calling journeys

adamweeks:


test/journeys/lib/test-helpers/space-widget/featureFlags.js
```diff
@@ -0,0 +1,48 @@
...
+      browserLocal.click(mainElements.exitButton);
+    });
+
+    it(`does not have a call option for user without flag`, () => {
+      browserRemote.click(mainElements.menuButton);
+      browserRemote.waitForVisible(mainElements.activityMenu);
+      assert.isFalse(browserRemote.element(mainElements.controlsContainer).element(meetElements.callButton).isVisible());
+      browserRemote.click(mainElements.exitButton);
+    });
+  });
```
bzang: May want to rename `browserRemote` and `browserLocal` to something more obvious to say that one has the feature flag enabled and one does not. Also does it make more sense to put the feature flag tests with the feature tests instead of here?

test/journeys/lib/test-helpers/space-widget/meet.js
```diff
@@ -4,6 +4,8 @@ import {switchToMeet} from './main';
 import {getEventLog} from '../../events';
 import {constructHydraId} from '../../hydra';
 
+export const FEATURE_FLAG_GROUP_CALLING = `js-widgets-group-calling`;
```
bzang: With more and more feature flags, should we create a package that just has the list of feature flags?

test/journeys/specs/space/dataApi/basic.js
```diff
@@ -81,6 +97,11 @@ describe(`Widget Space`, () => {
         biff.spark = new CiscoSpark({
           credentials: {
             authorization: biff.token
+          },
+          config: {
+            logger: {
+              level: `error`
+            }
```
bzang: YES! Thank jeebus!

test/journeys/lib/test-helpers/space-widget/featureFlags.js
adamweeks: * Good call on the renaming. 
* My thought on keeping the "has feature with flag, doesn't have it without it" tests were that we can keep track of all the feature flags in one test suite, while testing actual functionality in the others.

test/journeys/lib/test-helpers/space-widget/meet.js
adamweeks: We probably should come up with some way to track feature flags. If it is a regular package, we need to make sure it is available on npm since the journey tests don't get from the node modules though.

bzang: On thing I just realized this morning. You could reference packages just using relative paths... lot of ‘../../../‘ but that would work

PR #547, opened 2017-10-23
test(journeys): add message receipt to file xfer tests

adamweeks:
After watching all of the tests from today, sometimes the next file transfer will go across before the receipt message gets posted. Also, there was a weird assertion in send message that shouldn't be there.


bzang:
Nice catch!

PR #549, opened 2017-10-26
Group Call activity names

adamweeks:


packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -269,34 +273,66 @@ function getCallData(activity, currentUser) {
...
   const me = _.find(callObject.participants.items, (p) => p.person.entryUUID === currentUser.id);
 
+  const iWasJoined = me && me.state === `LEFT`;
   const iDeclined = me && me.state === `DECLINED`;
   const everyBodyJoined = _.every(callObject.participants.items, (p) => p.state === `LEFT`);
 
-  let arg;
-  let status;
+  const callees = _.filter(callObject.participants.items, (p) => !p.isInitiator);
+  const noBodyJoined = _.every(callees, (p) => p.state !== `LEFT`);
```
bzang: Can we import the lodash methods explicitly so that we can tree shake? Or use your suggestion to just use native methods.

adamweeks:
@bzang fixed up

PR #551, opened 2017-10-30
feat: Group calling event api

bzang:


packages/node_modules/@ciscospark/widget-meet/src/enhancers/call-memberships.js
```diff
@@ -0,0 +1,60 @@
+import {compose, lifecycle} from 'recompose';
+import {eventNames} from '../events';
+import {constructHydraId} from '@ciscospark/react-component-utils';
+
+export default compose(
+  lifecycle({
+    componentWillReceiveProps(nextProps) {
+      // Only add listeners if this is a new call
+      // eslint-disable-next-line no-warning-comments
+      // TODO: Need to create a better way to hook into call cycle
```
adamweeks: Ok, I'm sold, we should allow TODO comments :)

PR #552, opened 2017-10-31
Recents group calling

adamweeks:
https://voxeolabs.atlassian.net/browse/SSDK-1351

### DIFF
--- a/packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
+++ b/packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
@@ -122,6 +122,7 @@ export default function SpaceItem({ // eslint-disable-line complexity
         hasCallSupport &&
         <div className={classNames(`space-actions`, styles.actions)}>
           <Button
+            accessibilityLabel={`Call Space`}
             buttonClassName={classNames(styles.actionButton, styles.callButton)}
             iconType={ICON_TYPE_VIDEO_OUTLINE}
             onClick={handleCallClick}
--- a/packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
+++ b/packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
@@ -13,6 +13,7 @@ import styles from './styles.css';
 export default function SpaceItem({ // eslint-disable-line complexity
   avatarUrl,
   formatMessage,
+  hasGroupCalling,
   id,
   isUnread,
   lastActivityTime,
@@ -40,7 +41,7 @@ export default function SpaceItem({ // eslint-disable-line complexity
 
 
   function handleCallClick(e) {
-    if (type === `direct`) {
+    if (type === `direct` || hasGroupCalling) {
       e.stopPropagation();
       return onCallClick(id);
     }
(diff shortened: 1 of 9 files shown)


packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
```diff
@@ -40,7 +41,7 @@ export default function SpaceItem({ // eslint-disable-line complexity
 
 
   function handleCallClick(e) {
-    if (type === `direct`) {
+    if (type === `direct` || hasGroupCalling) {
```
bzang: What do you think about moving this check to a higher level. I don't think the `SpaceItem` component should determine if it should have a button or not, but maybe a prop like `hasCallingSupport` passed down from the parent.

packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
```diff
@@ -119,9 +119,10 @@ export default function SpaceItem({ // eslint-disable-line complexity
         {lastActivityTime}
       </div>
       {
-        isDirectSpace && hasCallSupport &&
+        hasCallSupport &&
         <div className={classNames(`space-actions`, styles.actions)}>
           <Button
+            accessibilityLabel={`Call Space`}
```
bzang: Do we need to start putting these labels into `intl` wrappers so that eventually they can be translated?

packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -141,16 +143,18 @@ function sortByNewest(space) {
...
+  [getWidget, getUsers, getSpaces, getRecentSpacesWithAvatarUrl, getSpark, getMedia, getTeams, getFeatures],
+  (widget, user, spaces, spacesList, spark, media, teams, features) => {
     let lastActivityDate;
     if (spacesList && spacesList.count()) {
       lastActivityDate = spacesList.last().lastActivityTimestamp;
     }
+    const hasGroupCalling = features.getIn([`items`, FEATURE_GROUP_CALLING]);
     return { // eslint-disable-line max-params
-      widgetRecent: widget,
+      widgetStatus: widget.get(`status`).toJS(),
```
bzang: Good call on being more specific here. Does the `toJS()` force too many rerenders or is the shallow comparison working? Would it be more performant if we convert the `getWidget` function to directly grab `state.widgetRecents.get('status')`?

test/journeys/specs/recents/dataApi/basic.js
```diff
@@ -49,7 +56,9 @@ describe(`Widget Recents`, () => {
             }
           }
         });
-        return marty.spark.internal.mercury.connect();
+        return marty.spark.internal.device.register()
+        .then(() => marty.spark.internal.feature.setFeature(`developer`, FEATURE_FLAG_GROUP_CALLING, true))
```
bzang: Nitpick, since we don't have the eslint rule yet. Indent here and next line unless it's off because of the diff.

packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
adamweeks: Yep. I created a ticket to fix all of these. https://voxeolabs.atlassian.net/browse/SSDK-1464

packages/node_modules/@ciscospark/widget-recents/src/selector.js
adamweeks: I honestly don't know. Would love to get these selectors as performant as possible, but not certain the best approach. Right now, as a starting step, I'm just trying to remove the immutable conversion from the renderer and into the selector, hopefully allowing the memoization to help.

test/journeys/specs/recents/dataApi/basic.js
adamweeks: Fixed.

packages/node_modules/@ciscospark/widget-recents/src/components/space-item/index.js
adamweeks: Yea, we're going to need some sort of space status object for each space in the spaces list, especially once the "call in progress" feature is added. Definitely something to think about.

PR #557, opened 2017-11-07
Tap tests run against CDN

adamweeks:


test/journeys/specs/tap/widget-space/oneOnOne.js
```diff
@@ -52,48 +53,15 @@ describe(`Widget Space: One on One: TAP`, () => {
...
-    const title = browserLocal.getTitle();
-    assert.equal(title, `Widget Space Production Test`);
-  });
-
-  it(`loads the user's name`, () => {
-    local.browser.waitForVisible(`h1.ciscospark-title`);
-    local.browser.waitUntil(() => local.browser.getText(`h1.ciscospark-title`) !== `Loading...`);
-    assert.equal(local.browser.getText(`h1.ciscospark-title`), remote.displayName);
-  });
-
```
moriahmaney: Is there a reason for deleting these two tests? I don't see a check for correct title/display name in the `loginAndOpenWidget` function

test/journeys/specs/tap/widget-space/space.js
```diff
@@ -201,8 +182,8 @@ describe(`Widget Space: Group Space: TAP`, () => {
       verifyMessageReceipt(local, remote, lorraineText);
       verifyMessageReceipt(remote, local, lorraineText);
       const clientSendEvents = getEventLog(local.browser);
-      assert.include(clientSendEvents, `messages:created`, `has a message created event`);
-      assert.include(clientSendEvents, `rooms:unread`, `has an unread message event`);
+      assert.isTrue(clientSendEvents.some((event) => event.eventName === `messages:created`), `event was not seen`);
+      assert.isTrue(clientSendEvents.some((event) => event.eventName === `rooms:unread`), `event was not seen`);
```
moriahmaney: Consider adding which type of event to the error message (`messages: created event was not seen`) because there's a lot of assertions here

test/journeys/specs/tap/widget-recents/index.js
```diff
@@ -33,7 +34,7 @@ describe(`Widget Recents`, () => {
 
   before(`load browser`, () => {
     browserLocal
-      .url(`/production/recents.html`)
+      .url(`https://code.s4d.io/widget-recents/latest/demo/index.html`)
```
moriahmaney: Is there a reason we aren't setting baseUrl to https://code.s4d.io/ if `process.env.TAP` is true?

adamweeks: No. There's not a good reason. Good call.

test/journeys/specs/tap/widget-space/oneOnOne.js
adamweeks: I guess I could add them back, I didn't really see a good reason for keeping them. They were originally there for a confirmation that the browser was ready, but that is being handled in the login function now

moriahmaney: Ok, works for me. Don't imagine that there would be a bug where the display name is wrong that wouldn't be caught by the check against `Send message to ${remote.displayName}` anyways.

adamweeks:
Blocked until /production is updated

PR #573, opened 2017-11-22
Presence feature

adamweeks:
Apologies for the massive PR, there are two main things happening here:
* widget-message uses the presence module to subscribe to all participants in a convo
* PresenceAvatar container displays an avatar with presence fetched from the module

Most of the commits are implementations of the PresenceAvatar and updates to the snapshots of those components.


adamweeks:
Deploy preview ready!

Built with commit 6dced185271ed580ab5e5ea8ef1a5cadc16d865a

https://deploy-preview-573--hopeful-hamilton-4090b9.netlify.com

adamweeks:
@bzang ready for a review now

packages/node_modules/@ciscospark/react-component-presence-avatar/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "0.1.0",
+  "description": "Cisco Spark React Presence Avatar",
+  "main": "./cjs/index.js",
+  "src": "./src/index.js",
+  "module": "./es/index.js",
+  "keywords": [],
+  "contributors": [
+  "Adam Weeks <adweeks@cisco.com>",
+  "Bernie Zang <nzang@cisco.com>"
+],
```
bzang: Indentation seems off here

.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
bzang: I didn't think comments are valid in JSON

packages/node_modules/@ciscospark/react-component-presence-avatar/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+    else if (presenceStatus === `ooo`) {
+      avatarClasses.push(`ciscospark-presence-avatar-ooo`);
+      avatarClasses.push(styles.ooo);
+      icon = <Icon type={ICON_TYPE_PTO} />;
+    }
+    else if (presenceStatus === `dnd`) {
+      avatarClasses.push(`ciscospark-presence-avatar-dnd`);
+      avatarClasses.push(styles.dnd);
+      icon = <Icon type={ICON_TYPE_DND} />;
+    }
```
bzang: These `else if`s seem like a good candidate for a `switch case` statement

packages/node_modules/@ciscospark/react-component-presence-avatar/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  return (
+    <div className={classNames(avatarClasses)}>
+      {icon &&
+        <div className={classNames(`ciscospark-presence-avatar-icon-container`, styles.iconContainer)} >
+          <div
+            className={classNames(`ciscospark-presence-avatar-icon`, styles.icon)}
+            style={{fontSize: props.iconSize}}
```
bzang: Should we keep the same setup as we did in the Avatar component and have size be: `small`, `medium`, `large`? I definitely prefer that to allowing arbitrary avatar sizes. Will make it easier later to unilaterally change sizes when the new Spring designs come in.

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
```diff
@@ -4,13 +4,13 @@ import classNames from 'classnames';
...
-import Avatar from '@ciscospark/react-component-avatar';
+import PresenceAvatar from '@ciscospark/react-container-presence-avatar';
 
-export default function TitleBar({children, image, name}) {
+export default function TitleBar({avatarId, children, name}) {
   return (
     <div className={classNames(`ciscospark-title-bar`, styles.titleBar)}>
       <div className={classNames(`ciscospark-avatar-container`, styles.avatarContainer)}>
-        <Avatar image={image} name={name} />
+        <PresenceAvatar avatarId={avatarId} iconSize={`30px`} name={name} />
```
bzang: Title bar avatars could also be space avatars right? Should there be a check here to make sure we pick the most efficient avatar component?

packages/node_modules/@ciscospark/widget-message/src/selector.js
```diff
@@ -27,33 +24,22 @@ export const getMessageWidgetProps = createSelector(
...
-      return p;
-    });
+
+    propsObject.activities = activities.toJS();
+    propsObject.activityCount = activities.count();
+    propsObject.lastActivity = activities.last();
+    propsObject.firstActivity = activities.first();
+    propsObject.conversationId = conversation.get(`id`);
+    propsObject.participants = participants.toJS();
+    propsObject.space = space;
```
bzang: would this be better as a single constant set? e.g.
``` js
const propsObject = {
  activities: activities.toJS(),
  activityCount: activities.count(),
  ...
};
```

packages/node_modules/@ciscospark/redux-module-presence/src/actions.js
```diff
@@ -155,22 +105,18 @@ export function subscribeToUpdates(userIds, spark) {
...
-      resource: `subscriptions`,
-      body: {
-        subjects,
-        subscriptionTtl: 0,
-        includeStatus: true
-      }
-    })
+    spark.internal.presence.unsubscribe(subjects)
       .then(() => {
         spark.internal.mercury.off(`event:apheleia.subscription_update`);
```
bzang: Would this clear all listeners instead of just the subjects you want to unsubscribe from?

.vscode/launch.json
adamweeks: vscode made this file so I guess it works for them

packages/node_modules/@ciscospark/react-component-presence-avatar/src/index.js
adamweeks: I think we've removed every instance of the named sizes from all avatar implementations in favor of arbitrary sizes everywhere.

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
adamweeks: Space avatars get their avatar urls from the same store users do, it would just fail to find presence for the space id.

packages/node_modules/@ciscospark/widget-message/src/selector.js
adamweeks: Yea, that looks cleaner, not sure why I went this route except maybe debugging. Will change it.

PR #577, opened 2017-11-29
feat(react-container-read-receipts): typing participants priority

adamweeks:
add sort so typing participants get priority

### DIFF
--- a/packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
+++ b/packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
@@ -1,5 +1,6 @@
 /* eslint-disable max-params */
 import {createSelector} from 'reselect';
+import {orderBy} from 'lodash';
 
 const getActivities = (state) => state.conversation.get(`activities`);
 const getParticipants = (state) => state.conversation.get(`participants`);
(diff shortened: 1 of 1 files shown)


adamweeks:
Deploy preview ready!

Built with commit fd75c3c96c3313c00e6164c01c9cf4061a92b271

https://deploy-preview-577--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable max-params */
 import {createSelector} from 'reselect';
+import _ from 'lodash';
```
bzang: We should pull out only the methods we use instead of importing all of lodash.

packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
```diff
@@ -16,25 +17,28 @@ export const getReadReceipts = createSelector(
     const readParticipants = participants
     .filter((participant) =>
       participant.get(`id`) !== currentUser.id &&
-      participant.getIn([`roomProperties`, `lastSeenActivityUUID`]) === activity.id);
+      participant.getIn([`roomProperties`, `lastSeenActivityUUID`]) === activity.id)
+    .toJS();
 
-    const limitParticipants = readParticipants.slice(0, 10)
+    const mappedParticipants = readParticipants
```
bzang: Would it be easier to add a `.sortBy` to the immutable object instead of converting the full object into a js object and then processing it? Wouldn't need lodash at that point.

packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
adamweeks: Ugh, we need to go through and do this throughout the code base, I always search and find the `import _`

packages/node_modules/@ciscospark/react-container-read-receipts/src/selectors.js
adamweeks: I've got some more stuff coming that benefits from converting it to a JS object so I'm going to stick with lodash

PR #584, opened 2017-12-04
Read receipt group tooltip

adamweeks:


adamweeks:
Deploy preview ready!

Built with commit 8664b373283904749ed14f05bb64969d036b72fa

https://deploy-preview-584--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-component-badge/src/index.js
```diff
@@ -14,9 +19,24 @@ const propTypes = {
...
+function Badge({children, tooltip}) {
+  return (
+    <span className={classNames('ciscospark-badge', styles.badge)}>
+      { tooltip &&
+        <div className={classNames('ciscospark-badge-tooltip', styles.tooltip)}>
+          <div className={classNames('ciscospark-badge-text', styles.tooltipText)}>
+            {tooltip}
+          </div>
+        </div>
+      }
```
bzang: Should the tool tip be attached to the badge component? This component seems a bit overloaded with this here. Would it be better as a standalone wrapping component?

adamweeks: I tried making the tooltip a dynamic component. The only solution I really could come up with was something that accepted a "X" position for where the tooltip should start/center. I didn't like any of them so I just punted it for the time being

packages/node_modules/@ciscospark/react-component-badge/src/styles.css
```diff
@@ -1,9 +1,59 @@
 .badge {
-  height: 18px;
+  position: relative;
+  display: flex;
+  width: 35px;
```
bzang: Just noticed this static width. What happens when the badge contents get wider than 35px that?

packages/node_modules/@ciscospark/react-component-badge/src/index.js
bzang: What about something that isolates this code so that a developer can do something like:
``` js
<Tooltip content={tooltip}><Badge>+1234</Badge></Tooltip>
```
The internals would be the same as the current implementation.

I'm thinking about this from the perspective of others using our components. If someone implements the badge with tooltip now and we pull the tooltips out later, it would cause either a breaking change or we'd have to maintain additional code so that it doesn't break.

adamweeks: That's exactly what I was trying to do, but couldn't get it to position properly because the sizing and location is based off of the child content, which the parent doesn't know.

packages/node_modules/@ciscospark/react-component-badge/src/styles.css
adamweeks: img

PR #586, opened 2017-12-05
WIP: Migrate eslint config

bzang:


adamweeks:
Deploy preview ready!

Built with commit d9a0e9927df3f791483e499b1a0c54299dc8810d

https://deploy-preview-586--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -150,13 +150,13 @@ export class MeetWidget extends Component {
       this.props.checkWebRTCSupport(sparkInstance);
     }
     if (isWebRTCSupported) {
-      this.setup(this.props);
```
adamweeks: What rule stops us from using `this`?

bzang: The setup method doesn’t actually use this, so I moved it to a static method which means we can’t reference it with this.setup. Nothing stops us from using this.

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
```diff
@@ -23,7 +22,7 @@ function TitleBar({avatarId, children, name}) {
   return (
     <div className={classNames('ciscospark-title-bar', styles.titleBar)}>
       <div className={classNames('ciscospark-avatar-container', styles.avatarContainer)}>
-        <PresenceAvatar avatarId={avatarId} iconSize={'30px'} name={name} />
+        <PresenceAvatar avatarId={avatarId} iconSize="30px" name={name} />
```
adamweeks: The react rule doesn't require these strings any more?

packages/node_modules/@ciscospark/react-component-typing-avatar/src/index.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import PresenceAvatar from '@ciscospark/react-container-presence-avatar';
-import classNames from 'classnames';
 
-import Avatar from '@ciscospark/react-component-avatar';
```
adamweeks: Not sure how it let me keep that in

packages/node_modules/@ciscospark/react-container-read-receipts/src/index.js
```diff
@@ -37,13 +37,12 @@ export class ReadReceipts extends Component {
   render() {
     const {users, totalCount} = this.props.readReceipts;
     const readReceipts = users.map((user) =>
-      <TypingAvatar
+      (<TypingAvatar
```
adamweeks: What does wrapping in parens give us?

bzang: This was enabled https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-wrap-multilines.md. Gives better readability for multi-line jsx

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
bzang: Yea. That was an oversight on my part. I think we should enable it as a warn, but as is we tend to ignore the rule and get around it anyways. What do you think? Is there another way we can encourage or enforce not using raw strings when it’s displayed text?

packages/node_modules/@ciscospark/react-container-read-receipts/src/index.js
adamweeks: img

No more eslint disable!

packages/node_modules/@ciscospark/react-component-title-bar/src/index.js
adamweeks: I haven't ever gotten around it, but I do like the idea of it. Although now that I'm reading the rule: https://github.com/yannickcr/eslint-plugin-react/blob/HEAD/docs/rules/jsx-no-literals.md I think this is just for string within a component and not params.

Issue #122, opened 2017-04-04
Activity "read" status handling


adamweeks:
The web client uses mouse movement to detect when an item has been read. We should switch to that pattern.

PR #587, opened 2017-12-06
Mark space read improvements

adamweeks:
Added a few things that the web client has. 

* If the window isn't active when the new activity arrives, we won't mark it as read. (previously, if it was scrolled to the bottom, it would auto-ack no matter what the status of the widget/window was)
* Moving the mouse while scrolled to the bottom will mark as read

### DIFF
--- a/packages/node_modules/@ciscospark/widget-space/src/container.js
+++ b/packages/node_modules/@ciscospark/widget-space/src/container.js
@@ -79,6 +79,7 @@ const injectedPropTypes = {
 export const ownPropTypes = {
   customActivityTypes: PropTypes.object,
   initialActivity: PropTypes.string,
+  isAcknowledgementsDisabled: PropTypes.bool,
   muteNotifications: PropTypes.bool,
   spaceId: PropTypes.string,
   startCall: PropTypes.oneOfType([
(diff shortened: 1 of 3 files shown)


adamweeks:
Deploy preview ready!

Built with commit 8d0538e869300472908ea7731f4f7ba2b3515a4f

https://deploy-preview-587--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-space/src/container.js
```diff
@@ -79,6 +79,7 @@ const injectedPropTypes = {
 export const ownPropTypes = {
   customActivityTypes: PropTypes.object,
   initialActivity: PropTypes.string,
+  isAcknowledgementsDisabled: PropTypes.bool,
```
bzang: Should we call this `hasAcknowledgementsDisabled`? I feel it's better grammatically, and signals that "Acknowledgements" inside this component are disabled (e.g. `hasNotificationsMuted`)
I've always considered the `is` prefix to be something specific to this component that doesn't need another noun to tell you what it is (e.g. `isOff`, `isMuted`). What do you think?

adamweeks: Yea, makes sense!

adamweeks:
Closes #122

PR #589, opened 2017-12-11
feat(widget-base): update redux devtools implementation

adamweeks:
Utilizing the serialize immutable method for better view on the store with the browser plugin

https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize

### DIFF
--- a/packages/node_modules/@ciscospark/widget-base/src/store.js
+++ b/packages/node_modules/@ciscospark/widget-base/src/store.js
@@ -4,8 +4,17 @@ import {
   createStore
 } from 'redux';
 import thunk from 'redux-thunk';
+import Immutable from 'immutable';
 
-const devtools = window.devToolsExtension || (() => (noop) => noop);
+/* eslint-disable no-underscore-dangle */
+const composeEnhancers
+  = typeof window === 'object' &&
+  window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
+    ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
+      serialize: {
+        immutable: Immutable
+      }
+    }) : compose;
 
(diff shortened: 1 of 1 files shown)


adamweeks:
Deploy preview ready!

Built with commit 22f546cba7df3f2d233f496f5a8d8ecc9023d334

https://deploy-preview-589--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-base/src/store.js
```diff
@@ -4,8 +4,17 @@ import {
...
-const devtools = window.devToolsExtension || (() => (noop) => noop);
+/* eslint-disable no-underscore-dangle */
+const composeEnhancers
+  = typeof window === 'object' &&
+  window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
+    ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
+      serialize: {
+        immutable: Immutable
+      }
+    }) : compose;
```
bzang: Should we add a production check here or is it OK to include in production builds?

adamweeks: https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f

Been seeing a lot of sites (including jira) using it production. I'm very happy to keep it that way.

bzang: That definitely makes sense. I'm mostly wondering about our e2e encryption and if this opens it up for attack. If not then I'm all for it. It also looks like there's ways to sanitize the data, so maybe that's one way we can go to give open access to devtools but still clean the data where we need to.

PR #591, opened 2017-12-11
fix(widget-recents): delete incoming call when connected

adamweeks:
we were listening to the wrong event


adamweeks:
Deploy preview ready!

Built with commit a5d867dee9f4dc541f9d0c30dc45ce89a47a5241

https://deploy-preview-591--hopeful-hamilton-4090b9.netlify.com

PR #594, opened 2017-12-13
Refactor widget base

bzang:
A big clean up of the base widget along with some refactoring of other related modules to be actual HOCs vs just helper functions. The base widget is now in `spark-widget-base`. The recents widget has been migrated to this, but the other widgets need some extra work to get to there due to a bug in the original `widget-base` package.

Still needed
- [x] Data API
- [x] Update Readme

PR #596, opened 2017-12-14
Migrate all widgets to new base

bzang:
Follow up to #594. The commits will clear up once that PR is merged.
This migrates the other widgets to the new base. Also a couple of small bug fixes and code cleanup


adamweeks:
Deploy preview ready!

Built with commit 0d4803c11e2df44af9668d992b64545fef1bf836

https://deploy-preview-596--hopeful-hamilton-4090b9.netlify.com

adamweeks:
Is there a way to go through just the changes? Was trying to compare commits, but they are different.

bzang:
@adamweeks, start with this commit: https://github.com/ciscospark/react-ciscospark/pull/596/commits/703fea9244b6d78194533fb022da9e1f1ae13f2d

Anything after is new. I found a couple of bugs that were fixed in this PR so I cherry picked them over to the refactor PR.

packages/node_modules/@ciscospark/widget-space/src/container.js
```diff
@@ -157,10 +157,6 @@ export class SpaceWidget extends Component {
     return nextProps !== this.props;
   }
 
-  componentWillUnmount() {
-    this.props.removeWidget();
```
adamweeks: Is this handled elsewhere? We still need to remove the widget on unmount.

bzang: So what the `.removeWidget` action did was clear the store. With the new base, that's not really an issue any more because of how we instantiate a new store. This was actually causing issues because if you try to compose this component with say a recents component inside the same app, it would wipe the complete store whenever the space component unmounted.

adamweeks: Gotcha, I was thinking of the meet widget that shut down the camera access on remove.

PR #602, opened 2017-12-19
feat(tooling): add firefox compatibility for journey tests

moriahmaney:


adamweeks:
Deploy preview ready!

Built with commit d5c4c929a4dcae4fd22315397f5e0a5e0e00de89

https://deploy-preview-602--hopeful-hamilton-4090b9.netlify.com

Jenkinsfile
```diff
@@ -113,11 +113,18 @@ ansiColor('xterm') {
...
              CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true npm test
-             CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true npm run test:automation -- --suite=oneOnOne
-             CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true npm run test:automation -- --suite=space
-             CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true npm run test:automation -- --suite=recents
-             CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true npm run test:automation -- --suite=multiple
+             CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true PORT=4567 SAUCE_CONNECT_PORT=5004 BROWSER=chrome npm run test:automation -- --suite=oneOnOne &
+             sleep 60 && CISCOSPARK_CLIENT_ID=C873b64d70536ed26df6d5f81e01dafccbd0a0af2e25323f7f69c7fe46a7be340 SAUCE=true PORT=4568 SAUCE_CONNECT_PORT=5005 BROWSER=chrome npm run test:automation -- --suite=space &
```
adamweeks: Is there something we could do instead of sleeping? Are we waiting for something to start before issuing the next command?

moriahmaney: This is the only workaround I could find to avoid sauce tunnels being started at the same time & crashing and burning on jenkins. AFAIK there's no other commands we could do here instead.

test/journeys/specs/multiple/index.js
```diff
@@ -7,6 +7,7 @@ import CiscoSpark from '@ciscospark/spark-core';
 import testUsers from '@ciscospark/test-helper-test-users';
 
 import waitForPromise from '../../lib/wait-for-promise';
+import {moveMouse} from '../../lib/test-helpers/index';
```
adamweeks: I would suggest one of two options here:
- use `import {moveMouse} from '../../lib/test-helpers';`
- rename to `mouse.js` and do `import {moveMouse} from '../../lib/test-helpers/mouse;`

There's no need to specify `index`

wdio.conf.js
```diff
@@ -1,14 +1,52 @@
...
+const argv = require('yargs').argv;
+
+const uuid = require('uuid');
+
+const {inject} = require('./scripts/tests/openh264');
+
+const browser = process.env.BROWSER || 'chrome';
+const tunnelId = uuid.v4();
+const port = process.env.PORT || 4567;
+const {suite} = argv;
```
adamweeks: If suite isn't defined, we should give it a default value via yargs. Maybe just "all"

test/journeys/specs/multiple/index.js
moriahmaney: oh, good point, i'll remove the `index`

Jenkinsfile
adamweeks: Is there a way to signal that the tunnel is setup in the wdio before hook somehow? My concern is that it make take +60 seconds to setup and then the job is failing.

wdio.conf.js
moriahmaney: good point, will add that!

Jenkinsfile
moriahmaney: I don't *think* so, because I believe the tunnel is started right when the webdriver process is started, before any hooks are executed. I may be wrong on this but I don't believe so.

moriahmaney: Also, its not that the tunnel needs to be fully setup (it seems), it just crashes when starting at exactly the same time. I've run at least 4-5 test runs that have been successful with this method. Maybe revisit if it becomes a problem?

adamweeks: I'm actually more concerned about unnecessarily extending the length of the tests than anything. Having them run it parallel should more than make up for it though! Just wanted to make sure we tried all the options available.

moriahmaney: Ah, gotcha. Right now, test runs are sitting at ~30 minutes, including build time, so I still think we're saving quite a lot of time even though we have to stagger the build since we doubled the test coverage.

adamweeks: img

PR #603, opened 2017-12-19
Implement SRI and signed hash

bzang:
This implements SRI generation along with hash signing. There are some doc updates as well associated with this.
Jenkins piece is at this point and may need some additional review.


adamweeks:
Deploy preview ready!

Built with commit 488ade380768a2d7aa841776fb64d853cadbc0fd

https://deploy-preview-603--hopeful-hamilton-4090b9.netlify.com

scripts/utils/sri.js
```diff
@@ -0,0 +1,120 @@
+const fs = require('fs');
+const path = require('path');
+const crypto = require('crypto');
+
+const glob = require('glob');
+const rimraf = require('rimraf');
+const sriToolbox = require('sri-toolbox');
+
+
+function generateSRI({
```
adamweeks: Would like to see jsdoc on all of these

README.md
```diff
@@ -60,6 +62,47 @@ Once a widget is bundled, the version number is available in the following ways:
...
+>     }
+>   ]
+> }
+> ```
+
+To find the SRI hash for a specific build of the widgets, you can use the following URL, replacing {VERSION} with the specific version you are looking for: `https://code.s4d.io/widget-space/archives/{VERSION}/manifest.json`.
+
+Additionally, for those who want an additional layer of verification, a `signature` has been provided that is signed by a private key. You can use our [public key](./public-key.pem) to verify this signature. Here is example of the verification process using the nodeJS built-in `cryto` library:
+
+``` js
```
adamweeks: I like the code sample in the readme, very nice!

bzang:
@adamweeks there's a script that runs from the jenkinsfile here: https://github.com/ciscospark/react-ciscospark/pull/603/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R152

They get generated into the `./dist` path. Should we also commit a manifest into the repo? I feel like that's not necessary since we don't actually store the bundled files in the repo.

bzang:
@adamweeks ready for another go.

Jenkinsfile
```diff
@@ -132,6 +132,7 @@ ansiColor('xterm') {
             sh '''#!/bin/bash -ex
             source ~/.nvm/nvm.sh
             nvm use v8.9.1
+            git diff
```
ianwremmel: are you planning to keep this once you've get the jenkins issue worked out?

bzang: I was planning on it. It'll give us some extra info if there's another of the same issue.

PR #604, opened 2017-12-21
chore(jenkins): use project npmrc file

adamweeks:
This change will append and reset the project's npmrc file instead of the home directory's one which was causing issues on jenkins.

### DIFF
--- a/Jenkinsfile.jssdk
+++ b/Jenkinsfile.jssdk
@@ -24,13 +24,11 @@ ansiColor('xterm') {
           withCredentials([
             string(credentialsId: 'NPM_TOKEN', variable: 'NPM_TOKEN')
           ]) {
-            sh 'echo \'//registry.npmjs.org/:_authToken=${NPM_TOKEN}\' > $HOME/.npmrc'
             sh '''#!/bin/bash -ex
             source ~/.nvm/nvm.sh
             nvm install v8.9.1
             nvm use v8.9.1
             npm install
-            rm -f $HOME/.npmrc
             npm run upgradespark
             '''
           }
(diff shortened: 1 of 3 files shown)


adamweeks:
Deploy preview ready!

Built with commit 0312a5f15ed886656170658ecb257351d6499a1d

https://deploy-preview-604--hopeful-hamilton-4090b9.netlify.com

Jenkinsfile.jssdk
```diff
@@ -24,13 +24,11 @@ ansiColor('xterm') {
           withCredentials([
             string(credentialsId: 'NPM_TOKEN', variable: 'NPM_TOKEN')
           ]) {
-            sh 'echo \'//registry.npmjs.org/:_authToken=${NPM_TOKEN}\' > $HOME/.npmrc'
             sh '''#!/bin/bash -ex
             source ~/.nvm/nvm.sh
             nvm install v8.9.1
             nvm use v8.9.1
             npm install
-            rm -f $HOME/.npmrc
```
bzang: Do we no longer need private repo access for updating spark version? If not, I think we can safely remove the withCredentials call from this block.

adamweeks: I was going to leave it in until the test users module is fully public.

bzang: That makes sense, but this change seems to indicate that we're removing the line that injects the token into `.npmrc`, so the `NPM_TOKEN` env var isn't used anywhere.

adamweeks: Oh, I see, yea, we probably don't need it then.

PR #605, opened 2018-01-02
Files activity

adamweeks:
This first version behaves like the web client and displays the share activities currently loaded. We have another ticket to enhance it to load via search endpoint that may not make it to this release.


adamweeks:
Deploy preview ready!

Built with commit c64ecbc44bf34ed1314b9ad5408e099e529864e0

https://deploy-preview-605--hopeful-hamilton-4090b9.netlify.com

bzang:
Still reviewing the code, but it seems that the demo doesn't allow you to download the file when clicking the download button. Also some errors in the console, including:
```
Failed prop type: Invalid prop `fileShares` of type `array` supplied to `t`, expected `object`.
    in t (created by Connect(t))
```
```
Each child in an array or iterator should have a unique "key" prop.
```

adamweeks:
@bzang will look into it, thanks.

adamweeks:
@bzang ok, those issues should be fixed now, ready for review

packages/node_modules/@ciscospark/react-component-activity-share-files/src/index.js
```diff
@@ -49,24 +49,14 @@ export function ActivityShareFiles(props) {
...
-      <ShareFile
-        file={file}
-        isPending={isPending}
-        key={file.url}
-        onDownloadClick={onDownloadClick}
+      <FileShareDisplay
+        {
+          ...{
+            file, isFetching, isPending, key, objectUrl, onDownloadClick
+          }
```
bzang: This style feels gross to me, but I don't know why. Could constructing a fileShareProps object be better?

packages/node_modules/@ciscospark/react-component-file-share-display/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+};
+
+function FileShareDisplay(props) {
+  const {
+    file,
+    isFetching,
+    isPending,
+    objectUrl,
+    onDownloadClick
+  } = props;
```
bzang: We should probably use destructuring here since the actual `props` object isn't used anywhere. i.e.:
``` js
function FileShareDisplay({
  file,
  isFetching,
  isPending,
  objectUrl,
  onDownloadClick
}) {
// blah blah blah
}
```

packages/node_modules/@ciscospark/react-component-file-share-display/src/index.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+const renderer = new ShallowRenderer();
+
+describe('FileShareDisplay component', () => {
+  let props;
+  const handleClick = jest.fn();
+
+  beforeEach(() => {
+    props = {
+      avatar: <div />,
```
bzang: Is this prop used for these tests?

packages/node_modules/@ciscospark/react-component-widget-header/src/index.js
```diff
@@ -12,31 +12,29 @@ import styles from './styles.css';
...
+  title: PropTypes.string
 };
 
 const defaultProps = {
-  participantsCount: 0
+  title: ''
 };
 
-function RosterHeader({
+function WidgetHeader({
```
bzang: Is `WidgetHeader` the appropriate name here? When I think `WidgetHeader` I think of the title bar that has the Space Name or the name of the person I'm talking to. What do you think about `ActivityHeader` or `ActivityMenuHeader`?

bzang:
One more thing I noticed. It seems files that don't have previews render weird in the file's sidebar. They seem to extend outside of the main view and a horizontal scrollbar appears.
screen shot 2018-01-03 at 10 37 45 am

packages/node_modules/@ciscospark/react-component-activity-share-files/src/index.js
adamweeks: I was attempting to implement [this](https://medium.com/@tevthuku/less-code-defining-your-react-props-eb093c7d3768) but I'm pretty sure it feels gross to me too, lol.

packages/node_modules/@ciscospark/react-component-file-share-display/src/index.test.js
adamweeks: copypastabusted.

packages/node_modules/@ciscospark/react-component-widget-header/src/index.js
adamweeks: Good call.

adamweeks:
Ok @bzang all those issues should be fixed now including the attachment sizing.

PR #607, opened 2018-01-03
fix(widget-message): activities are no longer an immutable object

adamweeks:
Apparently flagging messages was broken along with deleting messages. No more!


adamweeks:
Deploy preview ready!

Built with commit 77f461cf9c1326051fdc55f07de248fde86c5a1f

https://deploy-preview-607--hopeful-hamilton-4090b9.netlify.com

PR #609, opened 2018-01-04
Scroll lock behavior

adamweeks:
Should follow this:
## Scrolling

* Lock scroll to bottom on initial load
* Scroll locks to the bottom unless user scrolls up
* once the user scrolls to bottom, it locks again

## Read

* activity is marked read when window has focus and is scrolled to the bottom
* if list manually scrolls to the bottom, mark as read even if not in focus (can scroll out of focus on mac)

### DIFF
--- a/packages/node_modules/@ciscospark/widget-message/src/actions.js
+++ b/packages/node_modules/@ciscospark/widget-message/src/actions.js
@@ -32,6 +32,44 @@ export function showScrollToBottomButton(isVisible) {
   };
 }
 
+/**
+ * Sets if the widget has been scrolled up from the bottom
+ *
+ * @export
+ * @param {boolean} isScrolledUp
+ * @returns {Thunk}
+ */
+export function setScrolledUp(isScrolledUp) {
+  return (dispatch, getState) => {
+    const {widgetMessage} = getState();
+    // Since we are triggering this every scroll, let's not attack
+    // our store if we don't need to
+    if (!isScrolledUp) {
+      /* eslint-disable operator-linebreak */
+      if (
+        widgetMessage.get('hasNewMessage') ||
+        widgetMessage.get('hasScrolledUp') ||
+        widgetMessage.get('showScrollToBottomButton')
+      ) {
+        dispatch(updateWidgetState({
+          hasNewMessage: false,
+          hasScrolledUp: false,
+          showScrollToBottomButton: false
+        }));
+      }
+    }
+    /* eslint-disable operator-linebreak */
+    else if (
+      !widgetMessage.get('hasScrolledUp') ||
+      !widgetMessage.get('showScrollToBottomButton')
+    ) {
+      dispatch(updateWidgetState({
+        hasScrolledUp: true,
+        showScrollToBottomButton: true
+      }));
+    }
+  };
+}
 
 export function updateHasNewMessage(hasNew) {
   return (dispatch) => {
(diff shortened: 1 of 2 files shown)


adamweeks:
Deploy preview ready!

Built with commit 74c18a07baebfbbeeee19d9ac80f28ed1a408a08

https://deploy-preview-609--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-message/src/actions.js
```diff
@@ -15,15 +14,6 @@ export function updateWidgetState(state) {
...
 
-export function setScrollPosition(scrollPosition) {
-  return {
-    type: SET_SCROLL_POSITION,
-    payload: {
-      scrollPosition
-    }
-  };
-}
-
```
bzang: I don't think we can get rid of scroll position. It's used to calculate where the scroll position should be set when new messages come in. 
Looking at the netlify demo, when I scroll to the top of the activity list and more messages load, the view jumps to the oldest message instead of staying in the same spot with the same messages in view.

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -262,22 +284,6 @@ export class MessageWidget extends Component {
...
-   * @param {number} prevScrollTop
-   * @returns {null}
-   */
-  updateScroll(firstActivity, previousFirstActivity, prevScrollTop) {
-    const {activityList} = this;
-
-    if (firstActivity && previousFirstActivity && firstActivity.id !== previousFirstActivity.id) {
-      activityList.setScrollTop(activityList.getScrollHeight() - this.scrollHeight + prevScrollTop);
-    }
-  }
```
bzang: Yea, can't get rid of this because it breaks message loading behavior.

adamweeks: Darn, I was hoping to simplify the scroll tracking behavior. That stinks.

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -608,22 +598,23 @@ export class MessageWidget extends Component {
...
-    else if (!widgetMessage.get('showScrollToBottomButton')) {
-      props.showScrollToBottomButton(true);
+    else {
+      props.setScrolledUp(true);
     }
 
     if (activityList.isScrolledToTop() && firstActivity.verb !== 'create') {
+      // Store scroll position before loading messages so the window
+      // doesn't jump after they load
+      props.setScrollPosition({scrollTop: activityList.getScrollTop()});
```
adamweeks: Scroll position is only used when loading previous activities and was causing weird jumps when scrolling near the bottom (~150px, where the `isScrolledToBottom` buffer comes into play), so I only save scroll position when we've scrolled up and start loading.

adamweeks:
@bzang ok, scroll positioning is back and "scroll up to load" should be working again.

adamweeks:
@bzang I changed it to throttle after reading [this](https://css-tricks.com/debouncing-throttling-explained-examples/#article-header-id-6) that was linked in the lodash docs

bzang:
Right, which is why I'm wondering if throttle is right. Do we need the guarantee of an execution everything X ms, or do we only want it after a user stops scrolling.

adamweeks:
@bzang understood. The most important part of handle scroll is to determine if have scrolled away from the bottom and then when we get to the top. We might be able to use debounce with a leading config, but I was happy with the performance that throttle got us without affecting overall app performance.

PR #613, opened 2018-01-08
Refactor users store

bzang:
Created new users module. Migrated recents module over to it. Need substantially more work to get the rest of the widgets over.


adamweeks:
Deploy preview ready!

Built with commit 424af0fc6703f39232f797514a0fb881e70a1963

https://deploy-preview-613--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
```diff
@@ -110,12 +114,15 @@ export function getSpaces(sparkInstance, options = {}) {
       items.forEach((space) => {
         if (!space.tags.includes(TAG_HIDDEN)) {
           spaces[space.id] = constructSpace(space, listOptions.deferDecrypt);
+          dispatch(storeUsers(space.participants.items));
```
adamweeks: Should this go in an `else` after `if (space.decrypt)`?

packages/node_modules/@ciscospark/redux-module-users/src/actions.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+  };
+}
+
+/**
+ * Retrieves the current user using internal APIs
+ * @param {Object} sparkInstance
+ * @returns {Function}
+ */
+export function fetchCurrentUser(sparkInstance) {
```
adamweeks: Let's get some snapshot tests on this module.

packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
bzang: This is should happen even if we don't enable deferred decrypt. When we do defer, this line adds the encrypted users as placeholders. If we don't defer, this line will add the already decrypted users.

adamweeks: Do we get a proper ID or email for encrypted users that will be updated after the decrypt?

bzang:
@adamweeks , ready for round 2

packages/node_modules/@ciscospark/redux-module-spaces/src/actions.test.js
```diff
@@ -153,12 +154,7 @@ describe('redux-module-space actions ', () => {
   });
 
   it('has exported actions', () => {
-    expect(actions.fetchSpaces).toBeDefined();
-    expect(actions.getSpaces).toBeDefined();
-    expect(actions.fetchSpace).toBeDefined();
-    expect(actions.removeSpace).toBeDefined();
-    expect(actions.updateSpaceWithActivity).toBeDefined();
-    expect(actions.updateSpaceRead).toBeDefined();
+    expect(actions).toMatchSnapshot();
```
adamweeks: Oooh, I like that

PR #614, opened 2018-01-09
Media Module: Call objects as Records

adamweeks:
This is the first part of the recents call tracking feature, but broken into a more manageable PR. It does two things:
- Converts call objects in media module to records
- Uses media module for incoming call tracking in Recents widget


adamweeks:
Deploy preview ready!

Built with commit d282d1dae819cec18ce1520018e3f079fd18c77d

https://deploy-preview-614--hopeful-hamilton-4090b9.netlify.com

PR #619, opened 2018-01-12
Fix message parsing and conversion

adamweeks:
I have another ticket to switch to the commonmark module that the web client is now using.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-activity/src/actions.test.js
+++ b/packages/node_modules/@ciscospark/redux-module-activity/src/actions.test.js
@@ -132,7 +132,7 @@ describe('redux-module-activity actions', () => {
     it('can submit a post activity', () => {
       const store = mockStore({});
       const activity = initialState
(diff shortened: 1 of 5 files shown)


adamweeks:

Built with commit b73dac9c55a6bdc7b82394d7ef88ce577116e40c

https://deploy-preview-619--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/redux-module-activity/src/actions.test.js
```diff
@@ -168,7 +168,7 @@ describe('redux-module-activity actions', () => {
     it('saves a failed share activity', () => {
       const store = mockStore({});
       const activity = initialState
-        .set('text', 'I\'m walking here!')
+        .set('text', '**bold** yo!')
```
bzang: Instead of using the same text for every test, should we switch it up a bit so we can also test different types of markup to get better coverage?

PR #645, opened 2018-02-02
Ability to auth with Jwt token

adamweeks:


adamweeks:

Built with commit c1152c40dba4e483760664f0f980913859428347

https://deploy-preview-645--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
```diff
@@ -24,71 +90,32 @@ export default function createSpark(accessToken, options = {}) {
...
+  let {ciscosparkInstance} = window;
+
+  if (typeof ciscosparkInstance === 'object') {
+    return Promise.resolve(ciscosparkInstance);
+  }
+
+  const credentials = JSON.parse(localStorage.getItem('credentials') || false);
+  if (credentials.access_token) {
+    return Promise.resolve(createSpark(credentials.access_token, options));
+  }
```
bzang: Are we sure we want to use a stored accessToken to authenticate even when someone passes a JWT? Is it possible for another user's credentials to get cached and when a JWT is passed, the old/different user credentials are used?

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -31,10 +31,16 @@ const injectedPropTypes = {
...
 const propTypes = {
-  accessToken: PropTypes.string.isRequired,
+  accessToken: PropTypes.string,
+  jwtToken: PropTypes.string,
   ...injectedPropTypes
 };
 
+const defaultProps = {
+  accessToken: '',
+  jwtToken: ''
```
bzang: do we want to use the term `jwtToken`. We should probably match the term used my the dev portal and hydra, which I think may still be something like `guestId`?

adamweeks: Good call, I'll change it to match the portal which calls it "Guest Token"

packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
adamweeks: Ah yea, good catch

adamweeks: Should we remove that feature from the regular access token login method too? Since we're using `LocalForageStoreAdapter` we probably shouldn't manually inspect localstorage at all for spark credentials

packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
```diff
@@ -12,7 +13,83 @@ import '@ciscospark/internal-plugin-team';
 import Spark from '@ciscospark/spark-core';
 import LocalForageStoreAdapter from '@ciscospark/storage-adapter-local-forage';
 
-export default function createSpark(accessToken, options = {}) {
+/**
+ * Creates the default spark config for the widgets
+ * @param {Object} [options]={}
```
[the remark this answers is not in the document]
adamweeks: Oh, it should be `[options={}]` to indicate default value, I'll fix it

adamweeks:
@bzang the 
>Just tried it out and I'm getting: Cannot read property 'requestAccessTokenFromJwt' of undefined

issue should be resolved

packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
```diff
@@ -24,71 +101,36 @@ export default function createSpark(accessToken, options = {}) {
...
+
+  if (typeof ciscosparkInstance === 'object') {
+    return Promise.resolve(ciscosparkInstance);
+  }
+
+  ciscosparkInstance = new Spark({
+    config: defaultConfig(options)
+  });
+  window.ciscosparkInstance = ciscosparkInstance;
+  return ciscosparkInstance.authorization.requestAccessTokenFromJwt({jwt}).then(() => ciscosparkInstance);
```
adamweeks: @ianwremmel this is the line I'd like to get your feedback on. Since we won't be passing an access token, is the instance that is returned after jwt auth the same thing that would be returned with an access token?

ianwremmel: yea, that'll work

PR #655, opened 2018-02-14
fix(redux-module-media): listen for declined incoming events

adamweeks:
Our recents widget will not hide the incoming call screen if the call is handled outside of the recents widget. This will handle the "decline from elsewhere" state. The "answer from elsewhere" state will need an sdk feature.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
+++ b/packages/node_modules/@ciscospark/redux-module-media/src/actions.js
@@ -3,6 +3,7 @@ import {constructCallState} from './helpers';
 export const UPDATE_STATUS = 'media/UPDATE_STATUS';
 
 export const DISMISS_INCOMING_CALL = 'media/DISMISS_INCOMING_CALL';
+export const ANSWERED_INCOMING_CALL = 'media/ANSWERED_INCOMING_CALL';
 export const STORE_CALL = 'media/STORE_CALL';
 export const CONNECT_CALL = 'media/CONNECT_CALL';
 export const REMOVE_CALL = 'media/REMOVE_CALL';
@@ -100,6 +101,15 @@ export function dismissIncomingCall(callId) {
   };
 }
 
+function answeredIncomingCall(callId) {
+  return {
+    type: ANSWERED_INCOMING_CALL,
+    payload: {
+      callId
+    }
+  };
+}
+
 export function hangupCall({call, locusUrl}) {
   return (dispatch) => {
     // Don't update call states after hangup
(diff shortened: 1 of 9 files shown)


adamweeks:

Built with commit 040927aa59f4d68adb111d80ee722f5ba67a3acd

https://deploy-preview-655--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -202,17 +202,29 @@ export function listenForIncomingCalls(sparkInstance, locusUrl) {
 
       return incomingCall.acknowledge()
         .then(() => {
-          const removeIncoming = () => dispatch(removeCall(incomingCall, incomingCall.locus.url));
+          const removeIncoming = () => {
+            incomingCall.off();
```
bzang: This seems to remove all listeners from the incoming call. Would this affect the bricklets in recents? i.e. when a call is rejected, it should still appear as an active call in the recents list.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -202,17 +202,29 @@ export function listenForIncomingCalls(sparkInstance, locusUrl) {
...
+          };
 
           // If the remote party hangs up before we accept/decline
           incomingCall.once('inactive', removeIncoming);
 
-          // Once we are connected, our connected call state will handle disconnects
+          // If we rejected the call elsewhere, remove it
+          incomingCall.on('membership:declined', () => {
+            if (incomingCall.me.state === 'rejected') {
+              incomingCall.off('inactive', removeIncoming);
```
bzang: Should we be keeping calls around until they become inactive? If remove them we can't maintain a list of active calls.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
adamweeks: The only time it calls `removeIncoming` is when the call goes inactive. We can probably remove the `.off` line if it is a concern, but it is immediately followed by the `removeCall` which removes it completely from our store anyway.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
adamweeks: True, instead of removing the call I'll do `declineIncomingCall` which marks the call as declined for recents.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
bzang: Gotcha, yea I may have read this before I saw the rest. I think it's more of a concern for the `membership:declined` event where we directly call `removeIncoming()`.

adamweeks:
@bzang updated, ready for round 2.

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -177,6 +177,37 @@ export function placeCall({
...
+ * Declines an incoming call
+ *
+ * @export
+ * @param {object} incomingCall
+ * @returns {Promise}
+ */
+export function declineIncomingCall(incomingCall) {
+  return (dispatch) => {
+    incomingCall.reject();
+    dispatch(removeCall(incomingCall));
```
bzang: Did we talk about not removing the call, but to flag it somehow?

adamweeks:
@bzang round three! good finds. Space and Meet widget will use the `isDismissed` flag to determine to display an incoming call like Recents does!

PR #656, opened 2018-02-14
chore(journeys): add coverage for chrome on windows 10

moriahmaney:
PR also contains commits for stability changes that were necessary to get all tests passing, and optimizations to reduce total runtime of the jenkins pipeline.


adamweeks:

Built with commit f987ec8b2b67add397eee70cc6195bfad3b6eded

https://deploy-preview-656--hopeful-hamilton-4090b9.netlify.com

package.json
```diff
@@ -20,6 +20,7 @@
     "test:automation:space": "wdio wdio.conf.js --suite space",
     "test:automation:recents": "wdio wdio.conf.js --suite recents",
     "test:tap": "cross-env TAP=true wdio wdio.conf.js --suite tap",
+    "test:integration": "wdio wdio.conf.js --suite recents ; sleep 60 ; wdio wdio.conf.js --suite multiple ; sleep 60 ; wdio wdio.conf.js --suite oneOnOne ; sleep 60 ; wdio wdio.conf.js --suite space",
```
adamweeks: Will the `sleep` command work on windows? Anything in the package scripts should be cross-platform.

test/journeys/lib/test-helpers/index.js
```diff
@@ -11,14 +11,14 @@ import uuid from 'uuid';
  * @returns {void}
  */
 // eslint-disable-next-line import/prefer-default-export
-export function moveMouse(aBrowser, selector, offsetX = 0, offsetY = 0) {
```
adamweeks: Bye 😢

test/journeys/specs/space/global/basic.js
```diff
@@ -121,16 +121,13 @@ describe('Widget Space', () => {
...
-      const spaceWidget = '.ciscospark-space-widget';
       browserLocal.execute((localAccessToken, spaceId) => {
         const options = {
           accessToken: localAccessToken,
           spaceId
         };
         window.openSpaceWidget(options);
       }, marty.token.access_token, conversation.id);
-
-      browserLocal.waitForVisible(spaceWidget);
```
adamweeks: We don't need to make sure the widget actually loads before continuing?

moriahmaney: @adamweeks the load is redundant as we have multiple other waits in each test. for some reason, having the waitForVisible in the before() block is what was causing the test time to explode in length

package.json
moriahmaney: good catch, let me double-check this and get back to you...

test/journeys/specs/space/global/basic.js
adamweeks: Cool, just making sure!

PR #661, opened 2018-02-21
Fix empty message submit

adamweeks:


adamweeks:

Built with commit 9912c6e715bc6700e47bb282da039ff0b7e7dc94

https://deploy-preview-661--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
```diff
@@ -134,6 +134,9 @@ export class MessageComposer extends Component {
       onSubmit
     } = props;
 
+    if (activity.get('text').length === 0 && activity.get('files').size === 0) {
+      return;
+    }
```
bzang: Does this check need to exist if the action has a check built in already?

adamweeks: I went ahead and doubled up just to be safe in case we used the action elsewhere

PR #668, opened 2018-02-28
Recents call progress

adamweeks:
Pretty massive PR, sorry. 

Major changes here:
* [recents selector gets any associated call with the space](https://github.com/ciscospark/react-ciscospark/commit/c7918a3599236a6a49adb7dd3120b8fd1a1cf574)
* [join call button added to space item if it has a call](https://github.com/ciscospark/react-ciscospark/commit/99134413397a06cd7b0a407b5cde4118f718e325)
* [`listenForCalls` method on media to listen for all calls created and incoming](https://github.com/ciscospark/react-ciscospark/commit/9f1aa995b1d694fc399c2cb7cdbc330d9c9ba507)


adamweeks:

Built with commit 97871d5d1c9e721821e3bde239b661c218a47f5e

https://deploy-preview-668--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/redux-module-media/src/actions.js
```diff
@@ -163,6 +163,62 @@ function bindEvents(dispatch, call) {
   call.on('error', (err) => dispatch(updateCallError(call, err)));
 }
 
+function handleIncomingCalls(sparkInstance, dispatch) {
```
bzang: The passing of dispatch always felt weird to me... I've generally turned things like this into thunks, but I can't really come up with a good reason besides "consistency"

packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -72,6 +75,11 @@ function constructGroup({space, team}) {
   return thisSpace;
 }
 
+const getIncomingCall = createSelector(
+  [getCalls],
+  (calls) => calls.find((call) => call.isIncoming && !call.isDismissed)
```
bzang: Do you mean to only get the first incoming call with `.find()`?

packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -118,55 +126,67 @@ const getRecentSpaces = createSelector(
...
           s.avatarUrl = avatarItems.get(s.id);
         }
-        return s;
-      });
-    }
-    return recentSpaces;
+      }
+      if (calls) {
+        s.call = calls.find((call) => {
+          const callSpaceId = call.instance.locus.conversationUrl.split('/').pop();
```
bzang: Instead of deeply searching for this, which could cause issues later when locus changes, why don't we add `conversationUrl` or the `spaceId` to the top level `media`? Or we can add `locusUrl` to the top level space Record since all Spaces have locus associated with them.

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -188,16 +206,28 @@ export class RecentsWidget extends Component {
...
+    const {
+      widgetStatus
+    } = props;
+    if (widgetStatus.hasFetchedCurrentCalls) {
+      return;
+    }
+    props.updateWidgetStatus({hasFetchedCurrentCalls: true});
+    props.checkCurrentCalls(props.sparkInstance);
+  }
+
```
bzang: Can this be added as an enhancer instead of into the component directly?

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -160,15 +173,20 @@ export class RecentsWidget extends Component {
...
-      }
-      if (isWebRTCSupported && !media.getIn(['status', 'isListening'])) {
-        props.listenForIncomingCalls(sparkInstance);
+        if (!hasCheckedWebRTCSupport) {
+          props.checkWebRTCSupport(sparkInstance);
+        }
+        if (isWebRTCSupported && !media.getIn(['status', 'isListening'])) {
+          RecentsWidget.getCurrentCalls(props);
+          props.listenForCalls(sparkInstance);
+        }
```
bzang: Could we move to the new mercury and media enhancer to get these checks? Would clean up the container a lot.

packages/node_modules/@ciscospark/widget-recents/src/components/space-item/styles.css
```diff
@@ -109,6 +109,10 @@
   white-space: nowrap;
 }
 
+.join-call {
+
+}
```
bzang: Empty styles?

adamweeks:
@bzang most of this work was done before the call enhancer was merged, so that is why it isn't taking advantage of it.

PR #670, opened 2018-02-28
fix(react-container-activity-list): timestamp does not update when minute changes

moriahmaney:


adamweeks:

Built with commit f5a523c79202d60570efdebe8a304443a62fd591

https://deploy-preview-670--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-container-activity-list/src/selectors.js
```diff
@@ -107,15 +107,20 @@ export const getActivityList = createSelector(
...
-    let lastActorId, lastDay, lastVerb;
+    let lastActorId, lastDay, lastMinute, lastVerb;
     let shouldDisplayNewMessageMarker = false;
 
     conversationActivities.forEach((activity) => {
       if (isActivityVisible(activity)) {
         // Insert day separator if this activity and last one happen on a different day
-        const activityMoment = moment(activity.published, moment.ISO_8601);
-        const activityDay = activityMoment.endOf('day');
+        const activityMomentDay = moment(activity.published, moment.ISO_8601);
```
adamweeks: Could you just do:
```js
const activityDay = moment(activity.published, moment.ISO_8601).endOf('day');
```
that way we won't have the confusion of moment changing its base object?

moriahmaney: Good point, I'll change that

PR #677, opened 2018-03-06
chore(package): update lagging packages

bzang:
Closes #530 #546 #562 #574 #583 #585 #648 #659 #667 #671 #676


bzang:
Tried updating to Webpack 4 with newest dev server and plugins but it looks like there are loaders including the ExtractTextPlugin and StyleLoader that aren't ready for Webpack 4 yet.

adamweeks:

Built with commit 5a8bef369368fe3dedf4f0fdcbca78b156306dc4

https://deploy-preview-677--hopeful-hamilton-4090b9.netlify.com

Jenkinsfile
```diff
@@ -89,72 +89,62 @@ ansiColor('xterm') {
...
+            nvm use v8.9.4
             npm run static-analysis
             '''
           }
 
           stage('Unit Tests') {
             sh '''#!/bin/bash -ex
-            source ~/.nvm/nvm.sh
-            nvm use v8.9.1
+            source ~/.nvm/nvm.sh &> /dev/null
```
adamweeks: Good!

package.json
```diff
@@ -11,20 +11,21 @@
...
-    "test:automation:recents": "wdio wdio.conf.js --suite recents",
-    "test:tap": "cross-env TAP=true wdio wdio.conf.js --suite tap",
-    "test:integration": "cross-env INTEGRATION=true wdio wdio.conf.js --suite integration",
+    "test:automation": "wdio ./wdio/wdio.conf.js",
+    "test:automation:oneOnOne": "wdio ./wdio/wdio.conf.js --suite oneOnOne",
+    "test:automation:space": "wdio ./wdio/wdio.conf.js --suite space",
+    "test:automation:recents": "wdio ./wdio/wdio.conf.js --suite recents",
+    "test:tap": "wdio ./wdio/wdio.tap.conf.js --suite tap",
+    "test:integration": "cross-env INTEGRATION=true wdio ./wdio/wdio.conf.js --suite integration",
+    "test:integration:sauce": "cross-env INTEGRATION=true wdio ./wdio/wdio.sauce.conf.js --suite integration",
```
adamweeks: needs `SAUCE=true` right?

adamweeks:
Would rather see the netlify/sauce stuff in its own PR, but if it depends on the updates, can we update the PR name?

bzang:
Yea. I pulled it in here to see if it could actually stabilize a build, which it definitely didn’t. This PR isn’t going to get merged as is. I’ll be taking out the sauce and netlify changes.

Jenkinsfile
bzang: So turns out that doesn't work. The `bash -x` ensures that everything gets logged.

package.json
bzang: I tried to remove all the flags that have logic into the `wdio.conf.js` to have specific files based on the the type of test or environment you're running in. That way we don't have to manage extra flags. `INTEGRATION` was one I couldn't get rid of because it's used in logic in our actual test suites. May want to move that out into a `pre` hook.

adamweeks:
Closing. Outdated. Reference only.

PR #679, opened 2018-03-07
Recents call progress and media updates

adamweeks:
Replaces #668


adamweeks:

Built with commit c45c83f2347ba57c7c00c0d98d7d829e5eca885f

https://deploy-preview-679--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -126,30 +126,36 @@ const getRecentSpaces = createSelector(
...
         }
         else {
           s.avatarUrl = avatarItems.get(s.id);
         }
-        return s;
-      });
-    }
-    return recentSpaces;
+      }
+      if (calls) {
```
adamweeks: @bzang I think you had a better idea on how to handle matching of calls to spaces?

bzang: Yea, if a `locus` object has a `conversationUrl` you can use that to match the call to a space.

bzang: Yea, the convo object actually has a `locusUrl` in it. We may need to alter the action helper and the space Record to actually store the value, but that should allow you to directly get the call instead of filtering.

adamweeks: Ok, I think we've got it mapped more efficiently now.

PR #682, opened 2018-03-08
chore: add standard-version

adamweeks:
We'll want to run this with #no-push once Jenkins is working again to make sure everything bumps properly, but it ran fine locally.

Our release commits didn't follow conventional commits, so they don't have proper CHANGELOG entries. (But they will after this)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -127,7 +127,7 @@ ansiColor('xterm') {
             source ~/.nvm/nvm.sh
             nvm use v8.9.1
             git diff
-            npm version patch -m "build %s"
+            npm run release -- --release-as patch --no-verify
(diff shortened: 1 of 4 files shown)


adamweeks:

Built with commit ed62d75012a0157d4ff187cb8b05a28bdca425ce

https://deploy-preview-682--hopeful-hamilton-4090b9.netlify.com

Jenkinsfile
```diff
@@ -127,7 +127,7 @@ ansiColor('xterm') {
             source ~/.nvm/nvm.sh
             nvm use v8.9.1
             git diff
-            npm version patch -m "build %s"
+            npm run release -- --release-as patch --no-verify
```
adamweeks: @bzang here is where we release as a patch version

PR #696, opened 2018-03-19
Static Postcss 2.1.1

adamweeks:


adamweeks:
https://github.com/postcss/postcss-loader/issues/347

package.json
```diff
@@ -157,10 +157,10 @@
     "mkdirp": "^0.5.1",
     "mockdate": "^2.0.1",
     "npm-check-updates": "^2.12.1",
-    "postcss": "^6.0.3",
+    "postcss": "^6.0.20",
```
bzang: Should we also set this as an exact version like `postcss-load`?

adamweeks:

Built with commit 3e52471ac5b0d6a08ee4c16dd8b5caa6f33d2bac

https://deploy-preview-696--hopeful-hamilton-4090b9.netlify.com

package.json
adamweeks: The loader is just the issue, we could probably remove this bump if we wanted to

PR #697, opened 2018-03-19
Guest Mode Journeys

adamweeks:


adamweeks:

Built with commit 3ead482c19936ca4c9cc9308f5849258480d3332

https://deploy-preview-697--hopeful-hamilton-4090b9.netlify.com

test/journeys/specs/oneOnOne/dataApi/guest.js
```diff
@@ -0,0 +1,116 @@
...
+  describe('Data API: Guest', () => {
+    let local, mccoy, remote, spock;
+
+    const mccoyName = 'Bones Mccoy';
+    const spockName = 'Mr Spock';
+
+    before('create test users', () => Promise.all([
+      setupTestUserJwt({displayName: spockName}).then((guestUser) => {
+        spock = guestUser;
+        local = {browser: browser.select('browserLocal'), displayName: spockName, user: guestUser};
```
bzang: This pattern seems a bit obfuscating-y (that's totally a word). Most of the helper functions only needed the browser and maybe one other piece of information. Instead of referencing this object (e.g. `local.browser`), can we instead use `browserLocal`?

We can move this line down to the 1 test that calls a method that still needs the full object?

test/journeys/specs/oneOnOne/global/guest.js
```diff
@@ -0,0 +1,122 @@
...
+  describe('Global: Guest', () => {
+    let local, mccoy, remote, spock;
+
+    const mccoyName = 'Bones Mccoy';
+    const spockName = 'Mr Spock';
+
+    before('create test users', () => Promise.all([
+      setupTestUserJwt({displayName: spockName}).then((guestUser) => {
+        spock = guestUser;
+        local = {browser: browser.select('browserLocal'), displayName: spockName, user: guestUser};
```
bzang: Similar to `dataApi` version

test/journeys/specs/oneOnOne/dataApi/guest.js
```diff
@@ -0,0 +1,116 @@
...
+        sendMessage(local, remote, message);
+        verifyMessageReceipt(remote, local, message);
+      });
+    });
+
+    describe('meet widget', () => {
+      describe('pre call experience', () => {
+        it('has a call button', () => {
+          switchToMeet(local.browser);
+          local.browser.element(meetElements.meetWidget).element(meetElements.callButton).waitForVisible();
```
bzang: Should avoid calls and directly call `waitForVisible()`

test/journeys/lib/test-users.js
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {Promise<object>}
+ */
+export default function setupTestUserJwt({displayName}) {
+  return createUser({displayName})
+    .then(({jwt}) => {
+      const guestSpark = new CiscoSpark();
+      return guestSpark.authorization.requestAccessTokenFromJwt({jwt}).then(() =>
+        // We don't have a user id for guest users until a record is looked up
+        guestSpark.people.get('me').then((p) => {
+          const guestUser = p;
```
bzang: Bypasses the eslint rule, but doesn't actually address the issue. We should use `Object.assign({}, p);` to copy the data instead of referencing it differently.

wdio.conf.js
```diff
@@ -1,8 +1,12 @@
-require('dotenv').config();
 require('babel-register');
-
 const os = require('os');
 
+const dotenv = require('dotenv');
+
+dotenv.config();
+dotenv.config({path: '.env.default'});
```
adamweeks: Added this to match the sdk's implementation of env vars

.env.default
```diff
@@ -0,0 +1 @@
+CISCOSPARK_APPID_ORGID=Y2lzY29zcGFyazovL3VzL09SR0FOSVpBVElPTi80MjFhMWY4OC0zM2ViLTQ3MTUtOTY1ZS0zYzE5ZDYyZDJjMDk
```
bzang: Should we keep the `APPID_ORGID` static here? I think it makes more sense to store both in the Jenkins config since one is useless without the other. If we need to change it, it would force us to do a PR instead of only making the change in Jenkins... unless that's the reason for splitting them up

test/journeys/specs/oneOnOne/dataApi/guest.js
```diff
@@ -0,0 +1,114 @@
...
+        const csmmDom = document.createElement('div');
+        csmmDom.setAttribute('class', 'ciscospark-widget');
+        csmmDom.setAttribute('data-toggle', 'ciscospark-space');
+        csmmDom.setAttribute('data-guest-token', localAccessToken);
+        csmmDom.setAttribute('data-to-person-email', localToUserEmail);
+        csmmDom.setAttribute('data-initial-activity', 'message');
+        document.getElementById('ciscospark-widget').appendChild(csmmDom);
+        window.loadBundle('/dist-space/bundle.js');
+      }, spock.jwt, mccoy.email);
+      browserLocal.waitForVisible(`[placeholder="Send a message to ${mccoy.displayName}"]`, 30000);
```
moriahmaney: Is there a reason to not just leave this as the default waitForVisible timeout?

test/journeys/specs/oneOnOne/dataApi/guest.js
```diff
@@ -0,0 +1,114 @@
...
+      });
+    });
+
+    describe('meet widget', () => {
+      describe('pre call experience', () => {
+        it('has a call button', () => {
+          switchToMeet(browserLocal);
+          browserLocal.waitUntil(
+            () => browserLocal.isVisible(`${meetElements.meetWidget} ${meetElements.callButton}`),
+            1500,
```
moriahmaney: These kinds of tiny timeout values are part of what caused so much flakiness in the refactor I just did; please increase or remove the custom timeout value.

test/journeys/specs/oneOnOne/global/guest.js
```diff
@@ -0,0 +1,120 @@
...
+      });
+    });
+
+    describe('meet widget', () => {
+      describe('pre call experience', () => {
+        it('has a call button', () => {
+          switchToMeet(browserLocal);
+          browserLocal.waitUntil(
+            () => browserLocal.isVisible(`${meetElements.meetWidget} ${meetElements.callButton}`),
+            1500,
```
moriahmaney: Same thing here with the timeout value.

.env.default
adamweeks: I'm simply matching the way the sdk utilizes the app id org field. We may need to change these in both places once @mattnorris's work is done

PR #700, opened 2018-03-20
Chore: move all tests to circle CI, refactor journey tests

bzang:
Ok, this is finally ready to go.

There's quite a bit, but it's basically applying the same pattern a few times.

There were a couple of places where I tried some optimization. If it works to stabilize things then we can move more tests over to this mode.


adamweeks:

Built with commit a9b6968e9ab2018369b5d70e61bd910a2e9c68a3

https://deploy-preview-700--hopeful-hamilton-4090b9.netlify.com

.huskyrc.yml
```diff
@@ -0,0 +1,3 @@
+hooks:
+  pre-commit: lint-staged && ./scripts/validate-ci-config.sh
```
adamweeks: Making this a bash script means you will not be able to commit from windows. Can it be converted to a node script instead?

.huskyrc.yml
```diff
@@ -0,0 +1,3 @@
+hooks:
```
adamweeks: Any reason for the change to the .yml file instead of leaving it in package.json? It no longer matches the way we do husky precommits in the js-sdk with this change.

package.json
```diff
@@ -147,15 +150,17 @@
     "extract-text-webpack-plugin": "^3.0.1",
     "file-loader": "^1.0.0",
     "html-webpack-plugin": "^2.30.1",
-    "husky": "^0.14.3",
+    "husky": "^0.15.0-rc.9",
```
adamweeks: Not a huge fan of having rc releases in. Looks like they did 3 rc versions in one day, currently at rc 12.

scripts/deploy/commands/netlify.js
```diff
@@ -0,0 +1,54 @@
+const path = require('path');
+const fs = require('fs');
+
+const netlify = require('netlify');
+require('dotenv').config();
+
+module.exports = {
+  command: 'netlify <deployDir> [options]',
```
adamweeks: awesome!

scripts/utils/tests/openh264.js
```diff
@@ -7,7 +7,7 @@ const denodeify = require('denodeify');
 const FirefoxProfile = require('firefox-profile');
 const {stat} = require('fs-promise');
 
-const spawn = require('../utils/spawn');
+const spawn = require('../spawn');
```
adamweeks: img

scripts/validate-ci-config.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
adamweeks: Take a look at the lint-staged section of js-sdk package.json. We probably don't need to run this before every commit. We could just do the linting only when the circle config file changes.

.circleci/config.yml
```diff
@@ -0,0 +1,365 @@
...
+# Main Config
+version: 2
+jobs:
+  install:
+    docker:
+      - image: circleci/node:8.9.4
+    steps:
+      - checkout
+      - run: echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} >> .npmrc
+      - run: npm install
```
adamweeks: would using the new npm flag `ci` speed things up a bit here? I've never actually used it

wdio.conf.js
```diff
@@ -1,312 +0,0 @@
...
-
-
-const browser = process.env.BROWSER || 'chrome';
-const platform = process.env.PLATFORM || 'mac 10.12';
-const tunnelId = uuid.v4();
-const port = process.env.PORT || 4567;
-const {suite} = argv || 'integration';
-const screenResolutionMac = '1920x1440';
-const screenResolutionWin = '1920x1080';
-const screenResolution = platform.toLowerCase().includes('os x') || platform === 'darwin' || platform.includes('mac') ? screenResolutionMac : screenResolutionWin;
```
adamweeks: This was not brought over to the new wdio

Jenkinsfile
```diff
@@ -77,84 +77,16 @@ ansiColor('xterm') {
...
-
-          stage('Unit Tests') {
-            sh '''#!/bin/bash -ex
-            source ~/.nvm/nvm.sh
-            nvm use v8.9.1
-            npm run jest
-            '''
-          }
-
-          stage('Journey Tests') {
```
adamweeks: img

test/journeys/lib/test-helpers/index.js
```diff
@@ -36,3 +48,148 @@ export function moveMouse(aBrowser, selector) {
...
+      config: {
+        logger: {
+          level: 'error'
+        }
+      }
+    });
+    return users[key];
+  }
+
+  for (let i = 0; i < count; i += 1) {
```
adamweeks: If `count` is greater than `userKeys`, does it handle properly with a generic username from the test user library?

test/journeys/lib/test-helpers/index.js
```diff
@@ -36,3 +48,148 @@ export function moveMouse(aBrowser, selector) {
...
+export function setupOneOnOneUsers() {
+  return createTestUsers(2, {
+    spock: {displayName: 'Mr Spock'},
+    mccoy: {displayName: 'Bones Mccoy'}
+  });
+}
+
+/**
+ * Loads a widget into the browser with Data API
+ * @param {Object} options
```
adamweeks: want to list the possible options here?

.huskyrc.yml
bzang: New version of husky is forcing a new format that doesn't use the `scripts` field in `package.json`. When I updated I figure it made sense to move that out of `package.json` since the file was getting pretty huge.

package.json
bzang: Yea, good call. I'll revert the husky changes.

.huskyrc.yml
bzang: Good call. I'll convert it into a node script, but the validator uses a CLI tool that isn't available on windows, so what I'll do is check if the command exists and the platform is windows before running the validator.

.circleci/config.yml
bzang: It takes about 1:08 to do a full install, right now, which I don't think is bad. The new `ci` flag needs a newer version of `npm` so I didn't put it in here yet.

Usually what I'd be able to do is restore from cache and then do `npm install` which would drop the install time down to about :30, but `npm` has a bug where it sometimes fails the install because it can't delete an old dependency correctly. Figure it's fine this way for now, and we can optimize later.

wdio.conf.js
bzang: Good catch

test/journeys/lib/test-helpers/index.js
bzang: Yup

test/journeys/lib/test-helpers/index.js
bzang: Yes

wdio/wdio.sauce.conf.js
```diff
@@ -16,26 +14,6 @@ if (process.env.BROWSER && process.env.BROWSER.includes('firefox')) {
   config.services.push('firefox-profile');
 }
 
-const baseOnPrepare = config.onPrepare;
```
adamweeks: Why are we removing the check?

test/journeys/specs/space/dataApi/startup-settings.js
```diff
@@ -0,0 +1,226 @@
...
+
+    before('load browsers', () => {
+      browser
+        .url('/data-api/space.html')
+        .execute(() => {
+          localStorage.clear();
+        });
+    });
+
+    before('create marty', () => testUsers.create({count: 1, config: {displayName: 'Marty McFly'}})
```
adamweeks: We could use your new helper functions here, correct?

wdio/wdio.conf.js
```diff
@@ -18,57 +19,45 @@ const SELENIUM_VERSION = '3.4.0'; // Do not update to 3.6.0. Breaks actions API
...
-else if (browserType.toLowerCase().includes('firefox')) {
-  browserCapabilities = {
-    browserLocal: {
-      desiredCapabilities: firefoxCapabilities
-    },
-    browserRemote: {
-      desiredCapabilities: firefoxCapabilities
-    }
-  };
+function getBrowserCapabilities(type = 'chrome', count = 2) {
```
adamweeks: Nifty! And this could allow us to run just one browser suites?

test/journeys/specs/space/dataApi/startup-settings.js
bzang: yea, just a spot i didn't touch because it wasn't erroring out regularly.

wdio/wdio.sauce.conf.js
bzang: Because collisions this doesn't actually prevent over loading sauce. In many cases another job would start on sauce between the time the check happens and we start a job. Need another way to do this that's on the Sauce side.

wdio/wdio.conf.js
```diff
@@ -123,6 +123,9 @@ exports.config = {
     'recents-multiple': [
       './test/journeys/specs/recents/**/*.js',
       './test/journeys/specs/multiple/**/*.js'
+    ],
+    test: [
```
adamweeks: Still need this?

.huskyrc.yml
adamweeks: Gotcha, we'll need to make that change to js-sdk's husky config too

wdio/wdio.sauce.conf.js
adamweeks: Gotcha, thanks

test/journeys/specs/space/dataApi/startup-settings.js
adamweeks: you forgot "yet", lol

bzang:
Superseded by #708 and #709

PR #705, opened 2018-03-23
fix(redux-module-media): Prevent overwriting call record on connect

bzang:
We were overwriting the Call record whenever we fired CONNECT_CALL with a normal object. this creating some very odd states when components couldn't actually access specific data using immutable methods and would fail silently.


adamweeks:

Built with commit 8a3b7ccd6411bcaacc15ff6dd8dc86b7c4a2bcd8

https://deploy-preview-705--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -71,7 +71,7 @@ export class MeetWidget extends Component {
               isActive={isActive}
               localVideoPosition={localVideoPosition}
               onHangupClick={handleHangup}
-              {...call}
+              {...call.toJS()}
```
adamweeks: I've been trying to keep all of the immutable conversions out of containers/renders. Could this go in the selector instead?

bzang: Unfortunately not without a big refactor. There are HOCs that need the call object to still be a immutable record and not a plain js object.

adamweeks: Ah, ok, and this worked previously because the call object was no longer an immutable object (which is the bug you fixed).

PR #713, opened 2018-04-16
chore(Jenkinsfile): git checkout reordering

adamweeks:
Tags were being cleared and never refetched.
Causing the release process not to be able to find previous version.

This way matches the order the js-sdk works.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,8 +46,10 @@ ansiColor('xterm') {
             sshagent(['d8533977-c4c5-4e2b-938d-ae7fcbe27aac']) {
               // return the exit code because we don't care about failures
               sh script: 'git remote add upstream git@github.com:ciscospark/react-ciscospark.git', returnStatus: true
-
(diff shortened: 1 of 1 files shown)


adamweeks:

Built with commit 13e8ef28ed9c722bee13f71a00b120576f610e20

https://deploy-preview-713--hopeful-hamilton-4090b9.netlify.com

Jenkinsfile
```diff
@@ -57,8 +59,6 @@ ansiColor('xterm') {
             }
 
             sh 'git checkout upstream/master'
-            sh 'git reset --hard && git clean -f'
```
bzang: Should we keep this in to make sure we're guaranteed a clean folder?

adamweeks: I'm hoping the `git gc` above does most of that for us, but let's keep an eye on it. If it does need a real clean and reset, we'll want to add it to the sdk as well.

PR #739, opened 2018-04-25
Fix 1:1 conversation title

mattnorris:
- Uses ID instead of email for 1:1 conversations
- Adds test
- Removes unused function 

Resolves https://jira-eng-gpk2.cisco.com/jira/browse/SSDK-1851

### DIFF
--- a/packages/node_modules/@ciscospark/widget-space/src/selector.js
+++ b/packages/node_modules/@ciscospark/widget-space/src/selector.js
@@ -10,6 +10,15 @@ const getCurrentUser = (state, ownProps) => ownProps.currentUser;
 const getFeatures = (state) => state.features;
 const getOwnProps = (state, ownProps) => ownProps;
 
+function getToUser(conversation, currentUser) {
+  // Return a CASE-INSENSITIVE filter
+  function getToUserFilter(participant) {
+    return participant.get('emailAddress').toLowerCase() !== currentUser.email.toLowerCase();
+  }
+
+  return conversation.get('participants').filter(getToUserFilter).first();
+}
+
 export const getSpaceDetails = createSelector(
   [getWidget, getConversation, getUsers, getCurrentUser],
   (widget, conversation, users, currentUser) => {
@@ -20,8 +29,7 @@ export const getSpaceDetails = createSelector(
       let toUser;
       // Grab avatar if we're in a direct conversation
       if (spaceDetails.get('type') === 'direct') {
-        toUser = conversation.get('participants')
-          .filter((p) => p.get('emailAddress') !== currentUser.email).first();
+        toUser = getToUser(conversation, currentUser);
         if (toUser) {
           title = toUser.get('displayName');
           avatarId = toUser.get('id');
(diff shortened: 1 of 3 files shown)


adamweeks:

Built with commit ca47b1f0bbdadbc48022ef350984cecc9b327f1b

https://deploy-preview-739--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-space/src/selector.js
```diff
@@ -5,23 +5,36 @@ const getWidget = (state) => state.widgetSpace;
...
+ * @param {participant} currentUser
+ * @returns {participant}
+ */
+export const getToUser = createSelector(
+  [getConversation, getCurrentUser],
+  /* eslint arrow-body-style: [0] */
+  (conversation, currentUser) => {
+    return conversation
+      .get('participants')
+      .filter((participant) => participant.get('id') !== currentUser.id)
```
adamweeks: `.find` would do the same thing as `.filter().first()`

packages/node_modules/@ciscospark/widget-space/src/selector.test.js
```diff
@@ -0,0 +1,54 @@
+import {fromJS} from 'immutable';
```
adamweeks: Hooray! Our first selector test!

packages/node_modules/@ciscospark/widget-space/src/selector.js
```diff
@@ -5,23 +5,36 @@ const getWidget = (state) => state.widgetSpace;
...
+ * Get the other user in a 1:1 conversation.
+ * @param {conversation} conversation
+ * @param {participant} currentUser
+ * @returns {participant}
+ */
+export const getToUser = createSelector(
+  [getConversation, getCurrentUser],
+  /* eslint arrow-body-style: [0] */
+  (conversation, currentUser) => {
+    return conversation
```
bzang: can we fix this style error and remove the eslint change?

packages/node_modules/@ciscospark/widget-space/src/selector.js
mattnorris: I like the efficiency. I will update.

packages/node_modules/@ciscospark/widget-space/src/selector.js
mattnorris: All the other functions are styled that way. I can refactor though.

mattnorris: Refactored. The slight difference was the appearance of any lines of code above the `return` statement. If you have one, e.g., assignment, it's fine. If you have `return` first, it wants you to use the single line syntax.

bzang: @mattnorris you should be able to do something like this:
``` js
(conversation, currentUser) => conversation
  .get('participants')
```
or 
``` js
(conversation, currentUser) =>
  conversation
    .get('participants')
```

PR #748, opened 2018-04-30
Recents components

adamweeks:


adamweeks:

Built with commit 840ab18817c357b1693bb0e27a59faec93f327a5

https://deploy-preview-748--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-component-space-item/src/index.js
```diff
@@ -34,13 +31,13 @@ const propTypes = {
...
 const defaultProps = {
+  activityText: '',
   avatarUrl: '',
-  call: undefined,
+  callStartTime: undefined,
   hasGroupCalling: false,
   id: '',
   isDecrypting: false,
   isUnread: false,
-  latestActivity: {},
```
bzang: Nice! Good clean up of the props!

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -537,6 +594,12 @@ export class RecentsWidget extends Component {
     const isWebRTCSupported = media.getIn(['webRTC', 'isSupported']);
 
     if (spacesList && hasFetchedSpaces) {
+      const spacesListWithActivityText = spacesList.map((space) => {
+        const s = Object.assign({}, space);
+        s.activityText = this.constructActivityText(space);
+        return s;
+      });
```
bzang: Seems like this, along with the `constructActivityText` are prime examples of code that can be a selector. What do you think?

packages/node_modules/@ciscospark/react-component-spaces-list/src/index.js
```diff
@@ -2,22 +2,42 @@ import React from 'react';
...
+  }).isRequired,
   formatMessage: PropTypes.func.isRequired,
   hasGroupCalling: PropTypes.bool,
   onCallClick: PropTypes.func,
   onClick: PropTypes.func,
-  spaces: PropTypes.object
+  spaces: PropTypes.oneOfType([
+    PropTypes.arrayOf(
+      PropTypes.shape({
+        avatarUrl: PropTypes.string,
```
bzang: What do you think about importing the propType for a space from the `spaceItem` component? On one hand, it's less repeated code. On the other, it does somewhat go against our goal of being more explicit in our declarations.

packages/node_modules/@ciscospark/widget-recents/src/container.js
adamweeks: I started off making this a selector, but it renders a component and uses format message so I kept it in the main render process. Would like for it to not have to re-process every time, so I'm up for suggestions on how to do those within a selector. Maybe we can just memoize it somehow outside of the render.

PR #749, opened 2018-04-30
Fix NPM static analysis

mattnorris:
It looks like packages may have been locked on the last run. This copies the config into the home directory to ensure the project config is maintained before adding the new token.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,7 +79,7 @@ ansiColor('xterm') {
           }
 
           stage('Clean') {
-           sh 'rm -rf node_modules'
+            sh 'rm -rf node_modules'
           }
 
           stage('Install') {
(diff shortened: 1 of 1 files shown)


adamweeks:

Built with commit 1aed8401d2c9552e809fa1cfbf1c15bf1309fbb6

https://deploy-preview-749--hopeful-hamilton-4090b9.netlify.com

Jenkinsfile
```diff
@@ -207,7 +207,10 @@ ansiColor('xterm') {
                 string(credentialsId: 'NPM_TOKEN', variable: 'NPM_TOKEN')
               ]) {
                 try {
-                  sh 'echo \'//registry.npmjs.org/:_authToken=${NPM_TOKEN}\' > $HOME/.npmrc'
+                  // Copy & update config file
+                  sh 'cp .npmrc $HOME/.npmrc'
```
adamweeks: Did we want to backup the existing npmrc in the home directory if it exists since other teams might be relying on it?

mattnorris: The web team confirmed that they weren't. I'd rather have each team be responsible for setting up and tearing down the environments properly than perpetuate dependence on files that may not be there.

adamweeks: Sound good to me. LGTM!

PR #750, opened 2018-05-01
Update CircleCI config to v2 and add journey tests

bzang:


adamweeks:

Built with commit 2304ef7a741200fd0738f7a30bc01871e4a5fdcc

https://deploy-preview-750--hopeful-hamilton-4090b9.netlify.com

scripts/build/commands/test.js
```diff
@@ -0,0 +1,42 @@
+const path = require('path');
+
+const fse = require('fs-extra');
+const rimraf = require('rimraf');
+
+const {exec} = require('../../utils/exec');
+
+module.exports = {
+  command: 'test <distPath>',
+  desc: 'Bundle the main widgets into a distributable folder for testing',
```
adamweeks: Not a fan of the name `test` for this command. Maybe `test-bundle` or something else?

.npmrc
```diff
@@ -1 +0,0 @@
-package-lock=false
```
adamweeks: Still not sure how this will work with greenkeeper. I had to implement a new script to get greenkeeper to update the package lock in a different project. Could we remove package locks from this PR or is it required?

bzang: Removing it makes it a little harder to guarantee exact reproducibility in builds, but for now I think it should be fine. The cache is keyed off of the package lock file, but I should be able to just key it off of package.json.

scripts/build/commands/test.js
bzang: 👍 That makes sense

PR #757, opened 2018-05-07
feat(space-item): add active prop

andrewholsted:
This adds an activeSpaceId prop to the `react-component-spaces-list` and `react-component-space-item`. The idea being you pass it into the list itself and each space item checks if it's "active" or not. When it is, it adds an `.active` class to the component. 

Question:

I've never really used snapshots before. I wanted to explicitly assert that the `active` class was added to the component (since the first time my snapshot rendered, it wasn't). Is there a better way to do that?

Todo: Add tests for the spaces-list component? There's none presently.

### DIFF
--- a/packages/node_modules/@ciscospark/react-component-space-item/src/__snapshots__/index.test.js.snap
+++ b/packages/node_modules/@ciscospark/react-component-space-item/src/__snapshots__/index.test.js.snap
@@ -1,10 +1,64 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`SpaceItem component correctly renders an active space 1`] = `
+<div
+  className="space-item item active"
+  onKeyDown={[Function]}
+  onMouseDown={[Function]}
+  role="button"
+  tabIndex="0"
+>
+  <div
+    className="space-avatar-wrapper avatarWrapper"
+  >
+    <div
+      className="space-avatar-border"
+    >
+      <Avatar
+        baseColor="blue"
+        image=""
+        isSelfAvatar={false}
+        name="Webex User"
+      />
+    </div>
+  </div>
+  <div
+    className="space-item-meta meta"
+  >
+    <div
+      className="space-team-name teamName"
+      style={
+        Object {
+          "color": "#f5f5f5",
+        }
+      }
+    >
+      Best Team
+    </div>
+    <div
+      className="space-title title"
+    >
+      Webex User
+    </div>
+    <div
+      className="space-last-activity lastActivity"
+    >
+      I'm active!
+    </div>
+  </div>
+  <div
+    className="space-last-activity-time timestamp"
+  >
+    3:05 PM
+  </div>
+</div>
+`;
+
 exports[`SpaceItem component renders a decrypting space 1`] = `
 <div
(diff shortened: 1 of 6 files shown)


packages/node_modules/@ciscospark/react-component-space-item/src/__snapshots__/index.test.js.snap
```diff
@@ -1,5 +1,59 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`SpaceItem component correctly renders an active space 1`] = `
+<div
+  className="space-item item active"
```
adamweeks: Since the snapshot file has the `active` class, we don't need to verify it in the actual test.

packages/node_modules/@ciscospark/react-component-space-item/src/index.test.js
```diff
@@ -44,6 +44,31 @@ describe('SpaceItem component', () => {
...
+        activeSpaceId="active-space-id"
+        onCallClick={onClick}
+        onClick={onClick}
+        name="Webex User"
+        teamColor="blue"
+        teamName="Best Team"
+      />
+    );
+    const component = renderer.getRenderOutput();
+    console.log(component.props.className)
```
adamweeks: We'll need to remove this console.

packages/node_modules/@ciscospark/react-component-space-item/src/index.test.js
```diff
@@ -44,6 +44,31 @@ describe('SpaceItem component', () => {
...
+    expect(component.props.className.includes('active')).toBeTruthy();
```
adamweeks: This is the test we don't need due to the snapshot being valid.

packages/node_modules/@ciscospark/react-component-space-item/src/index.js
```diff
@@ -91,7 +94,8 @@ function SpaceItem({
     <div
       className={classNames('space-item', styles.item, {
         [styles.hasCallSupport]: !!hasCallSupport,
-        [styles.isDecrypting]: !!isDecrypting
+        [styles.isDecrypting]: !!isDecrypting,
+        [styles.active]: activeSpaceId === id
```
adamweeks: I think it might make more sense to change this prop to `isActive` and do the `activeSpaceId` comparison in the spaces list.

andrewholsted: That works. Will update.

packages/node_modules/@ciscospark/react-component-space-item/src/index.test.js
andrewholsted: When I originally created the snapshot, it didn't have the `active` class. So my snapshot test was passing but it didn't really have everything I wanted. How do you know that the snapshot represents what you're expecting? Just manually validate and once you do that, you know it's good and you keep the snapshot?

adamweeks: Gotcha, yea, when a snapshot is initially created, you have to manually verify it looks like you want it to.

PR #769, opened 2018-06-04
Dynamically load avatars

moriahmaney:


adamweeks:

Built with commit 7049b46e8176e0768eeaacf15266bff702c6c4aa

https://deploy-preview-769--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-roster/src/components/add-participant/index.js
```diff
@@ -74,7 +74,7 @@ function AddParticipant({
     }
     else if (searchResults.results && searchResults.results.length) {
       results = (
-        <div className={classNames(`${baseClassPrefix}-results`, styles.results)}>
+        <div style={{height: '100%', width: '100%'}} className={classNames(`${baseClassPrefix}-results`, styles.results)}>
```
bzang: Should these inline styles be added to the css file instead?

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
```diff
@@ -1,6 +1,9 @@
+import 'react-virtualized/styles.css';
+
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import {List, AutoSizer} from 'react-virtualized';
```
bzang: How much bigger does virtualized make this component? If it's significant, this change should probably be localized to the widget since not every time a people list is used virtualizating is necessary.

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
```diff
@@ -16,56 +19,99 @@ const propTypes = {
...
+      totalRows.push({type: 'person', details: person});
+    });
+  });
+
+  function rowRenderer(options) {
+    const {
+      key,
+      index,
+      style
+    } = options; {
```
bzang: This should probably be destructured in the function param like this:
```js
function rowRenderer({key, index, style}) {
```

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
```diff
@@ -16,56 +19,99 @@ const propTypes = {
...
-          key={p.id}
-          onClick={handleItemClick}
-          {...p}
-        />
+        <div key={key} style={style}>
+          <Person
+            displayName={person.displayName || person.name}
+            key={person.id}
+            onClick={handleItemClick}
+            {...person}
```
bzang: Instead of spread syntax, we should reference the exact params we're passing in.

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -90,14 +90,14 @@ export class MessageWidget extends Component {
    */
   static getAvatarsFromParticipants(props) {
     const {
-      participants,
       sparkInstance
     } = props;
+    const participants = filter(props.activities, 'actor').map((activity) => activity.actor.id);
```
bzang: Much cleaner!

packages/node_modules/@ciscospark/widget-roster/src/container.js
```diff
@@ -153,7 +154,7 @@ export class RosterWidget extends Component {
       let content;
       if (currentView === VIEW_ADD) {
         content = (
-          <div>
+          <div style={{height: '100%', width: '100%'}}>
```
bzang: Move inline styles to css? 3 more below this

packages/node_modules/@ciscospark/widget-roster/src/components/add-participant/index.js
adamweeks: The autosizer for virtualized doesn't seem to like actual css styles, but only react ones. I had to do the same for recents.

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
adamweeks: The amount of work making two different virtualized vs non-virtualized scrolling components was rather large since @moriahmaney is handling the avatar loading only during scrolling.

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
adamweeks: FYI, this is a copy pasta from my code which was a copy pasta from virtualized demo docs. Is there an eslint rule we can add for these?

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
```diff
@@ -16,56 +19,99 @@ const propTypes = {
...
+          />
+        </div>
       );
-    });
+    }
+  }
+
+  function onRowsRendered({startIndex, stopIndex}) {
+    const userIds = totalRows.slice(startIndex, stopIndex).filter((row) => row.type === 'person').map((p) => p.details.id);
```
adamweeks: Could you throw a quick comment on what this is doing?

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -90,14 +90,14 @@ export class MessageWidget extends Component {
    */
   static getAvatarsFromParticipants(props) {
```
adamweeks: Should we rename this to something like `getAvatarsFromActivityActors`? or just `FromActors`?

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
moriahmaney: yepp, will do

packages/node_modules/@ciscospark/widget-message/src/container.js
moriahmaney: yeah, i meant to rename it but got distracted. have a preference between those two?

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
moriahmaney: will do

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
```diff
@@ -16,56 +19,99 @@ const propTypes = {
...
-            }
-            {group.people && group.people.length && generateGroupList(group.people)}
-          </div>
-        ))
+        <AutoSizer>
+          {({height, width}) => (
+            <List
+              height={height}
+              rowCount={totalRows.length}
+              rowHeight={44}
```
adamweeks: Are the header rows the same size as the people cells? If not, we'll have to make this prop a function to test if it is a header or person and return the correct size

moriahmaney: yeah, I tried this but the issue is the totalRows array is undefined at the point where the row height gets defined. I'll keep looking in to it today.

packages/node_modules/@ciscospark/react-component-people-list/src/index.js
```diff
@@ -16,56 +19,108 @@ const propTypes = {
...
+    if (item.label) {
+      totalRows.push({type: 'header', details: item.label});
+    }
+    // Rows for each person
+    item.people.forEach((person) => {
+      totalRows.push({type: 'person', details: person});
+    });
+  });
 
+  // disable react/prop-types because eslint thinks rowRenderer is the main render
```
adamweeks: img

PR #770, opened 2018-06-05
Refactor Activity Share with Selector

adamweeks:
While refactoring the activity share to use a selector, I found there were some missing icons around downloading and file sharing as well.

This should fix both. Commits should be proper.


adamweeks:

Built with commit 7e449606271bff513cae695b8d954614c1156991

https://deploy-preview-770--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-container-file-downloader/src/selectors.js
```diff
@@ -0,0 +1,16 @@
+import {createSelector} from 'reselect';
+
+const getSpark = (state, ownProps) => ownProps.sparkInstance || state.spark.get('spark');
+const getShare = (state) => state.share;
+
+const getContainerProps = createSelector(
```
mattnorris: Should this have a corresponding selector test?

PR #771, opened 2018-06-05
refactor(file-uploader): use reselect to get data from store

adamweeks:


adamweeks:

Built with commit 7818749b5ef72eeb9617ed3298613c27ae6c08cb

https://deploy-preview-771--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/react-container-file-uploader/src/selectors.js
```diff
@@ -0,0 +1,18 @@
+import {createSelector} from 'reselect';
+
+const getSpark = (state, ownProps) => ownProps.sparkInstance || state.spark.get('spark');
+const getConversation = (state) => state.conversation;
+const getActivity = (state) => state.activity;
+
+const getContainerProps = createSelector(
```
mattnorris: Should we create a selector test for this? Or let the existing tests cover this path?

adamweeks: There really isn't anything to test on this, but we could. It really just helps memoize the parts of the store that we want for the container.

PR #777, opened 2018-06-15
feat(widget-space): allow default overrides in activity menu

moriahmaney:


adamweeks:

Built with commit 183d804b29e5058f236b313937a45362d613f260

https://deploy-preview-777--hopeful-hamilton-4090b9.netlify.com

test/journeys/specs/oneOnOne/global/startup-settings.js
```diff
@@ -0,0 +1,115 @@
+import {assert} from 'chai';
```
adamweeks: img

test/journeys/testplan.md
```diff
@@ -92,6 +92,9 @@ The "oneOnOne" test suite opens a space widget between two individuals. The test
       - during call experience
         - can hangup in call
         - can decline an incoming call
+  - Startup Settings
```
adamweeks: Thanks for updating these!

packages/node_modules/@ciscospark/widget-demo/src/components/demo-widget/index.js
```diff
@@ -134,6 +141,13 @@ class DemoWidget extends Component {
     return this.setState({mode: value, destinationId: ''});
   }
 
+  @autobind
+  handleActivitiesChange(event, isInputChecked) {
+    const activities = {...this.state.activities};
```
bzang: This can be written more succinctly like this: `const {activities} = this.state;`

packages/node_modules/@ciscospark/widget-space/src/enhancers/errors.js
```diff
@@ -90,6 +93,18 @@ function checkForErrors(props) {
...
+  activityTypes.forEach((activity) => {
+    if (activity.name === initialActivity && spaceActivities && spaceActivities[activity.name] === false
+    && !errors.get('errors').has(disabledInitialErrorId)) {
+      props.addError({
+        id: disabledInitialErrorId,
+        displayTitle: formatMessage(messages.disabledInitialActivity),
+        temporary: false
+      });
+    }
+  });
```
bzang: Are we sure we want to to acknowledge that the activity is disabled? I would think if the developer has purposely disabled the activity, we should say the activity provided is not valid.
I'm drawing a comparison from if you're not logged in and try to hit a private URL on github, it gives you 404 instead of a permission denied error.

moriahmaney: Ok, that works. So to make sure I understand what you mean -- just reword the error message?

packages/node_modules/@ciscospark/widget-demo/src/components/demo-widget/index.js
moriahmaney: Got it, thanks!

packages/node_modules/@ciscospark/widget-space/src/enhancers/errors.js
moriahmaney: Let me know if this works for you instead @bzang: 'Error: The selected initial activity is invalid'

bzang: Yea, I think it would make sense to change the wording, but also handle the case where if the initialActivity doesn't exist in our activity list we also pop the same message.

packages/node_modules/@ciscospark/widget-space/src/container.js
```diff
@@ -102,6 +114,8 @@ export class SpaceWidget extends Component {
       );
     }
     if (sparkInstance && spaceDetails) {
+      // Filter out activities that are disabled using the spaceActivities prop
+      activityTypes = activityTypes.filter((a) => spaceActivities[a.name] !== false);
```
bzang: Sorry, just noticed this. This filter should probably be placed in a selector so that it doesn't run every time the widget renders.

moriahmaney: Will do!

PR #778, opened 2018-06-20
feat(widget-space-demo): add option to disable default activities

moriahmaney:
The demo for disabling activities was added to widget-demo, which is not available on the CDN. This PR will make those options appear on the CDN space widget demo.

### DIFF
--- a/packages/node_modules/@ciscospark/widget-space/src/enhancers/errors.js
+++ b/packages/node_modules/@ciscospark/widget-space/src/enhancers/errors.js
@@ -94,7 +94,9 @@ function checkForErrors(props) {
   }
 
   const invalidActivityId = 'ciscospark.container.space.error.invalidActivity';
-  const validActivity = initialActivity ? activityTypes.some((activity) => activity.name === initialActivity) : true;
+  const defaultActivity = 'message';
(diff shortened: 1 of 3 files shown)


adamweeks:

Built with commit 28c9b40d5b10b3193ff01b30a08aecd9a293824d

https://deploy-preview-778--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-space/src/enhancers/errors.js
```diff
@@ -94,7 +94,9 @@ function checkForErrors(props) {
   }
 
   const invalidActivityId = 'ciscospark.container.space.error.invalidActivity';
-  const validActivity = initialActivity ? activityTypes.some((activity) => activity.name === initialActivity) : true;
+  const defaultActivity = 'message';
```
adamweeks: this change doesn't really fit into the commit message or PR. intentional?

moriahmaney: whoops, i shouldve split them. it's a bug fix i noticed because the space widget demo doesn't set an initial activity, so i just fixed the bug to check for the default activity instead of adding an initial activity to the space demo.

PR #780, opened 2018-06-25
feat(widget-space): destination id and type props

adamweeks:
Apologies for the massive PR, attempted to do commit by commit.

PR #781, opened 2018-06-26
Reset widget on destination prop change

adamweeks:
Builds on #780 and adds the ability to reload the widget


adamweeks:

Built with commit 70d42fabad2af5c4ab1b67f060f5bd4c3e4bc906

https://deploy-preview-781--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-space/src/enhancers/setup.js
```diff
@@ -106,6 +109,27 @@ function setup(props) {
...
+    }
+
+    // Check for destination Change
+    if (destination && prevProps.destination) {
+      const calculatedDestination = getDestinationFromProps(props);
+      const calculatedPreviousDestination = getDestinationFromProps(prevProps);
+      if (
+        calculatedPreviousDestination.id !== calculatedDestination.id
+        || calculatedPreviousDestination.type !== calculatedDestination.type
+      ) {
```
bzang: Should we also take into account if someone passes an email, and then passes in the `spaceId` of the same space? Instead of comparing current to previous props could we, instead, compare current props, to what we have in the space/conversation store?
That would also make our migration to a centralized spaces store a lot easier.

packages/node_modules/@ciscospark/widget-message/src/container.js
```diff
@@ -330,11 +338,12 @@ export class MessageWidget extends Component {
    * Attempts to fetch and establish a conversation
    *
    * @param {object} props
+   * @param {object} prevProps
    *
    * @memberOf MessageWidget
    *
    */
-  establishConversation(props) {
+  establishConversation(props, prevProps) {
```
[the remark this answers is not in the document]
adamweeks: I don't understand the concern here. Instead of adding autobinding to this function, I passed the props in from the call of the function. IMO `props` vs `this.props` in the function would be confusing for me to read.

packages/node_modules/@ciscospark/widget-space/src/enhancers/setup.js
adamweeks: Is this an over-optimization? Seems like a VERY rare case to code for. I'm not even sure how to go able getting started on implementing it.

bzang: Not really an over-optimization, more just looking forward to how this will be used when we're not using the single conversation store, but the spaces store. We should minimize server calls as much as possible to provide a responsive UI to the end user. If we have the data already, we shouldn't go and refetch it.

This accomplishes what we need for now.  We'll just eventually need check what's in the store and not only the previous prop to detect the need to fetch or to load from store.

adamweeks: @bzang I've been trying to figure out how to implement this in the space widget without relying on conversation service, but we still don't have a way to get a space via email. Any ideas?

bzang: One approach might be to get the "other" user in a convo (I think we have that in a selector somewhere). If you check if the convo is `one_on_one` and the other user's email matches, then that should be a match for the space to reuse.

Maybe create a single function that takes the email and conversation and determines that. We can then easily migrate it into the spaces module alter.

adamweeks: Ok, I see, let me give it a shot.

PR #782, opened 2018-06-27
Add support for @webex packages

adamweeks:
Tooling updates and an initial test with the loading screen.


adamweeks:

Built with commit d6df7022c25cf77216fd06c86c7b2284a0a72a0c

https://deploy-preview-782--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@webex/react-component-loading-screen/src/index.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+const defaultProps = {
+  loadingMessage: ''
+};
+
+function LoadingScreen({loadingMessage}) {
+  return (
+    <div className={classNames('ciscospark-loading', styles.loading)}>
+      <div className={classNames('ciscospark-spark-logo', styles.logo)} />
```
bzang: Should we rename classnames while we're migrating things over? I feel like if we don't do it now, it won't get done anytime soon.

adamweeks: Yea, good catch. `webex-teams` a good prefix?

adamweeks: Fixed

mattnorris: I may be missing something, but I still see `ciscospark` for class names instead of `webex-teams`.

scripts/utils/package.js
```diff
@@ -4,28 +4,79 @@ const {statSync, readdirSync} = require('fs');
...
+  catch (err) {
+    return false;
+  }
   return false;
 }
 
+/**
+ * Get single packages full path
+ * @param  {string} pkg specific package name or full path
+ * @returns {string|boolean} full path string or false if failed
```
mattnorris: This works, but kind of nags at me because the return type can change. Can we return empty string so it will still be falsey?

packages/node_modules/@webex/react-component-loading-screen/src/index.js
adamweeks: Oh great, I fixed the wrong component... Stand by

adamweeks:
@bzang @mattnorris ready for another round

PR #785, opened 2018-06-28
fix(tooling): cannot run journey tests locally

moriahmaney:


adamweeks:

Built with commit b4d3dee48c8a804b87a8b36633f902e0dc105f97

https://deploy-preview-785--hopeful-hamilton-4090b9.netlify.com

wdio.conf.js
```diff
@@ -339,6 +326,16 @@ exports.config = {
 };
 
 if (process.env.SAUCE) {
+  const sauceCapabilities = Object.assign({}, chromeCapabilities, {
```
adamweeks: We need to handle firefox capabilities as well, correct?

moriahmaney: good catch

adamweeks:
Closed via #780

PR #788, opened 2018-07-10
chore(tooling): use webpack env command line feature

adamweeks:
Utilizes the [environment variable config](https://webpack.js.org/guides/environment-variables/) to provide a starting string for the css class name.

That way, if you run `npm build:package widget-space` it will name all css classes with `widget-space--[classname]--[hash]`.

### DIFF
--- a/scripts/utils/exec.js
+++ b/scripts/utils/exec.js
@@ -2,7 +2,7 @@
  * Borrowed from React Bootstrap
  * https://github.com/react-bootstrap/react-bootstrap
  */
-const processExecSync = require('child_process').execSync;
+const {shellSync} = require('execa');
 
 const processExec = require('child-process-promise').exec;
 require('colors');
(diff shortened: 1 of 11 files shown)


adamweeks:

Built with commit 856c124b743a4614151baa0279799c66d60d66fd

https://deploy-preview-788--hopeful-hamilton-4090b9.netlify.com

scripts/utils/exec.js
```diff
@@ -2,7 +2,7 @@
  * Borrowed from React Bootstrap
  * https://github.com/react-bootstrap/react-bootstrap
  */
-const processExecSync = require('child_process').execSync;
+const {shellSync} = require('execa');
```
adamweeks: We were using the `child_process` tool, because bootstrap was using it at the time, but they've moved on to `execa` and I'm just following suit

PR #794, opened 2018-07-30
feat(widget-recents): use new avatar design

moriahmaney:


adamweeks:

Built with commit 9a66bc4efb0ddd0feccf80049facba6969f01d4b

https://deploy-preview-794--hopeful-hamilton-4090b9.netlify.com

bzang:
@moriahmaney , looks like a new version of JSDom breaks jest and our unit tests. Here's the jest ticket with instructions on how to fix the issue: https://github.com/facebook/jest/issues/6766

packages/node_modules/@webex/react-component-space-item/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@webex/react-component-space-item",
+  "description": "Webex Teams React Space Item",
+  "main": "./cjs/index.js",
+  "src": "./src/index.js",
+  "module": "./es/index.js",
+  "keywords": [],
+  "contributors": [
+    "Adam Weeks <adweeks@cisco.com>",
+    "Bernie Zang <nzang@cisco.com>"
```
bzang: You should add your name here!

packages/node_modules/@webex/react-component-space-item/src/avatar.scss
```diff
@@ -0,0 +1,6 @@
+$brand-font-folder: "~@collab-ui/core/fonts";
+$icon-font-path: "~@collab-ui/icons/fonts";
+$images-path: "~@collab-ui/core/images";
+$docs-images-path: '~@collab-ui/core/docs/assets';
+
+@import "~@collab-ui/core/scss/components/avatar";
```
bzang: Neat. Glad you got this working!
Have you had a chance to test how well this works as a standalone npm component? I'm wondering if the final generated CSS is going to have a lot of duplicates styles as we move more components over to collab-ui

moriahmaney: @bzang can I have some clarification on how I should test it standalone? As far as the CSS goes, from the SCSS in collab-ui it's all pretty well separated by component, and from the compiled CSS I see only the avatar styles have been brought in, so I'm thinking we should be ok.

bzang: Awesome. I was a bit worried about styles getting duplicated around.

As for testing the component, the primary goal I have is having a useable standalone component outside of our tooling stack. Beyond building the component and then importing it into a basic react project outside of our repo, we haven't implemented a cleaner way to test the component. Have any ideas?

One thought is to import it into one of the demos since that's pretty standalone compared to one of the widgets. @adamweeks any thoughts?

adamweeks: You could add it to the `samples` to test how it operates outside of the widgets. That doesn't give us a true external view though, especially before it gets published.

PR #796, opened 2018-07-30
feat(widget-demo): add initial activity setting

moriahmaney:


adamweeks:

Built with commit 430611ee09987eb68f6b36c6b167c5408dfa4c6a

https://deploy-preview-796--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-space-demo/src/components/demo-widget-space/index.js
```diff
@@ -229,8 +237,30 @@ class DemoWidgetSpace extends Component {
...
+                    aria-label="Meet"
+                    label="Meet"
+                    value="meet"
+                  />
+                </RadioButtonGroup>
+              </div>
+
               <div>
                 <h3> Widget Activities </h3>
+                {displayActivityWarning && (<div> Warning: The initial activity must be selected </div>)}
```
adamweeks: One thing that would make this a bit more like the rest of our JSX code is to put the html on a new line. Not sure if we have an eslint rule possible for this or not.

adamweeks: Does `Warning: The initial activity is not enabled` make more sense? When I first read this error, I thought it was that the initial activity wasn't chosen (selected).

bzang: We should probably also put that string into a i18n constant.

moriahmaney: @adamweeks yeah that's probably a better way to phrase it and @bzang will do

adamweeks: @bzang I don't believe we have any intl support in the demos

bzang: Oops, sorry. I confused myself for a moment. Yea normal text for this is fine until we get the internationalization requirement.

packages/node_modules/@ciscospark/widget-space-demo/src/components/demo-widget-space/index.js
```diff
@@ -229,8 +237,33 @@ class DemoWidgetSpace extends Component {
...
+                    aria-label="Meet"
+                    label="Meet"
+                    value="meet"
+                  />
+                </RadioButtonGroup>
+              </div>
+
               <div>
                 <h3> Widget Activities </h3>
+                {
```
adamweeks: img

PR #798, opened 2018-08-01
fix(tooling): use proper shell result for git commands

adamweeks:
When we switched to the `execa` method of running commands, the previous `String()` method of interpreting the results no longer works.

Also removed the unused `diff` command, since we hardcode diff into the Jenkinsfile. If needed in the future, we can get the current version from the sdk package.


adamweeks:

Built with commit 6e2cdbdd497491af46586f282c5abd6f3420d74d

https://deploy-preview-798--hopeful-hamilton-4090b9.netlify.com

PR #799, opened 2018-08-02
Reduce conversation requirements

adamweeks:
This PR does a few things to reduce the requirement of having a conversation object to function properly.

* The space details (avatar and title) are fetched via the users store or the spaces store depending on destination type
* Whenever a call is placed to a destination, it stores the resulting locus url for that destination in order to cross reference the destination with the media object. This means we don't have to get the locus url from conversation when we don't have a conversation.

Fixes (I came across a couple of bugs that I fixed as well, can be pulled into its own PR if necessary):
* call timer in header was using an old prop from the media store that is no longer there
* only does the "auto switch" back to messaging after hanging up a call if the messaging widget is available

A good way to test this PR would be to load the widget with everything but "meet" disabled and try calling different types of destinations. This should put us in good shape to add SIP and PSTN as destination types for space widget. 

img


adamweeks:

Built with commit 3ed6af381b319cdb4f2d181490453ce4018abe60

https://deploy-preview-799--hopeful-hamilton-4090b9.netlify.com

packages/node_modules/@ciscospark/widget-space/src/selector.test.js
```diff
@@ -2,9 +2,67 @@ import {fromJS, Map} from 'immutable';
 
 import {CallRecord} from '@ciscospark/redux-module-media';
 
-import {getCall} from './selector';
+import {getCall, getSpaceDetails} from './selector';
+
+import {destinationTypes} from './';
 
 describe('widget-space selectors', () => {
+  describe('#getSpaceDetails selector', () => {
```
bzang: Awwww yea!

PR #806, opened 2018-08-10
fix(widget-recents): re-add missing event fields

bzang:
At some point during the recents list changes a few weeks back the spaces list reference got switched to raw spaces and the events were missing the fields needed to create full events. I added those back in and put in some event tests in the journey tests to at least check fields are not empty.

### DIFF
--- a/test/journeys/lib/events.js
+++ b/test/journeys/lib/events.js
@@ -37,7 +37,7 @@ export function getEventLog(myBrowser) {
  * @param {object} options
  * @param {array} options.events
  * @param {string} options.eventName
- * @returns {(array|boolean)}
+ * @returns {(array)}
  */
 export function findEventName({events, eventName}) {
   if (Array.isArray(events)) {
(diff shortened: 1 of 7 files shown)


test/journeys/lib/events.js
```diff
@@ -29,3 +29,19 @@ export function getEventLog(myBrowser) {
...
+/**
+ * Searches events object for matching event name
+ *
+ * @export
+ * @param {object} options
+ * @param {array} options.events
+ * @param {string} options.eventName
+ * @returns {(array|boolean)}
+ */
+export function findEventName({events, eventName}) {
```
adamweeks: Nice little helper!

test/journeys/lib/events.js
```diff
@@ -29,3 +29,19 @@ export function getEventLog(myBrowser) {
...
 }
+
+/**
+ * Searches events object for matching event name
+ *
+ * @export
+ * @param {object} options
+ * @param {array} options.events
+ * @param {string} options.eventName
+ * @returns {(array|boolean)}
```
adamweeks: This should always return an array, correct? When would it return boolean?

bzang: derp, forgot to update the jsdocs after I made some changes.

bzang:
@adamweeks, ready for another review

PR #808, opened 2018-08-13
filter update content activities from store

adamweeks:
Content update activities come back from the conversation api out of (published time) order, so we need to remove them from the store and not count them in our max date calculation.

(Most of this code was taken from the web client's handling of activities)


packages/node_modules/@ciscospark/redux-module-conversation/src/actions.js
```diff
@@ -330,7 +330,7 @@ export function loadPreviousMessages(conversationId, maxDate, spark) {
       conversationId,
       lastActivityFirst: true,
       limit: 20,
-      maxDate
+      maxDate: Date.parse(maxDate)
```
bzang: May want to add an input check here to make sure we get a real date back, and not call the API if we don't.

PR #813, opened 2018-08-22
fix(widget-space): cannot place call with direct room id

moriahmaney:


packages/node_modules/@ciscospark/widget-meet/src/container.js
```diff
@@ -55,10 +56,15 @@ export class MeetWidget extends Component {
...
-      displayName,
       localVideoPosition
     } = this.props;
     const {formatMessage} = intl;
+    const {avatarId} = this.props.spaceDetails;
+    let {displayName} = this.props;
+
+    if (!displayName) {
+      displayName = this.props.spaceDetails.title;
+    }
```
bzang: Would this be more performant if it was in the selector instead of here in the render?

packages/node_modules/@ciscospark/react-component-incoming-call/src/index.js
```diff
@@ -1,14 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import Avatar from '@ciscospark/react-component-avatar';
+import PresenceAvatar from '@ciscospark/react-container-presence-avatar';
```
adamweeks: I feel a bit weird about having a container inside a component, because it means this component will require a redux store. @bzang @moriahmaney any ideas on how we could do this while keeping the component pure?

bzang: Good catch. I think the "right way"™ to do this is probably to pass in the Avatar as a prop from a level up.

moriahmaney: ok, will look into this -- should note though that other components that use PresenceAvatar import the container

packages/node_modules/@ciscospark/widget-meet/src/container.js
moriahmaney: Good call, will update this logic

PR #816, opened 2018-08-23
Fix nvm/npm prefix incompatibility

mattnorris:
-  Fix 'nvm is not compatible with the npm config "prefix". This error can happen when this agent's NODE_JS_BUILDER alters the environment.
- Upgrade package.


mattnorris:
This merged but did not automatically close.

PR #821, opened 2018-09-04
fix(r-m-conversation): apply proper update content filter

adamweeks:
The `if` logic was a bit off and filtering too many conversation activities causing file attachments coming from the conversation server to not be shown.

There is also a commit for a quick demo fix. I can separate into a different PR if needed.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
+++ b/packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
@@ -50,8 +50,8 @@ function filterActivities(activities) {
   // to make sure received activities are contiguous, remove the update
   // activities with object.objectType `content`
   return activities.filter((a) =>
-    a.verb !== 'update' &&
-    a.object && a.object.objectType !== 'content');
+    !(a.verb === 'update' &&
+    a.object && a.object.objectType === 'content'));
 }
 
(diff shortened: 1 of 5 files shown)


packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.test.js
```diff
@@ -226,7 +226,38 @@ describe('redux module conversation reducer', () => {
...
             url: 'http://activity.url/abc-123',
             verb: 'update',
+            mock: true,
+            note: 'this should not be in snapshot'
+          }]
+        }
+      })).toMatchSnapshot();
+    });
+
+    it('should not filter content activities', () => {
```
adamweeks: Added test to verify bug doesn't happen again.

packages/node_modules/@ciscospark/redux-module-conversation/src/reducer.js
```diff
@@ -50,8 +50,8 @@ function filterActivities(activities) {
   // to make sure received activities are contiguous, remove the update
   // activities with object.objectType `content`
   return activities.filter((a) =>
-    a.verb !== 'update' &&
-    a.object && a.object.objectType !== 'content');
+    !(a.verb === 'update' &&
```
mattnorris: 👍 that's rough

PR #822, opened 2018-09-07
Add giphy activities support for conversation

adamweeks:
The shape of the giphy activity as received from the server needs a bit of a massage to get it to play friendly with the `download()` function.


moriahmaney:
200w

LGTM

PR #835, opened 2018-09-17
test(journeys): skip start call setting test

adamweeks:
This test fails due to it showing a call already in progress.
Until we can determine the cause (bug ticket SPARK-36716), skip the failing test.


moriahmaney:
200w-2

PR #836, opened 2018-09-17
docs(testplan): change testplan to new strategy

adamweeks:


test/journeys/testplan.md
```diff
@@ -1,20 +1,82 @@
 # Test Plan
 
+The Webex Teams Widgets have full integration (or "journey") tests performed to validate proper behavior.
+
+The widgets are tested via [Webdriver.io](https://webdriver.io) using devices powered by [Sauce Labs](https://saucelabs.com).
+
+Every code change pull request (PR) is tested and verified by [Circle CI](https://circleci.com) by running the smoke test suite on the supported browsers/operating systems. Before a PR is opened, it is the responsibility of the developer to run the full suite on the widget that has changes.
```
mattnorris: By "responsibility of the developer", do you mean we are relying on the developer? Or the developer shouldn't be surprised if the tests don't pass if they don't test themselves?

adamweeks: As in, we are going to add a PR template like the sdk has that requires you to say which test suite you ran locally. Our full suite has been far too unstable to rely on circle ci to run the entire thing without expecting a failure. Only the "smoke tests" will run on circle/jenkins for the PR.

PR #838, opened 2018-09-18
fix(circleci): error on multiline command doesn't fail build

moriahmaney:
This *should* fix the error with reporting a success on a sauce crash type error

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,7 @@ jobs:
           name: Integration Tests
           no_output_timeout: 40m
           command: |
-            set -m
+            set -em
             (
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -108,7 +108,7 @@ jobs:
           name: Integration Tests
           no_output_timeout: 40m
           command: |
-            set -m
+            set -em
```
adamweeks: What does this change do?

moriahmaney: makes sure errors are not being suppressed: https://support.circleci.com/hc/en-us/articles/115015733328-Tests-fail-but-job-finishes-successfully

PR #840, opened 2018-09-21
feat(widget-meet): use alert type from locus to display incoming call alerts

adamweeks:
If an incoming call does not have an alert type, it should not ring.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-media/src/helpers.js
+++ b/packages/node_modules/@ciscospark/redux-module-media/src/helpers.js
@@ -51,7 +51,7 @@ function constructCallStatus(call) {
       isInitiated = false;
       isConnected = false;
     }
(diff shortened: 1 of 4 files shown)


packages/node_modules/@ciscospark/redux-module-media/src/helpers.js
```diff
@@ -51,7 +51,7 @@ function constructCallStatus(call) {
       isInitiated = false;
       isConnected = false;
     }
-    isIncoming = call.direction === 'in' && !call.hasJoinedOnThisDevice;
+    isIncoming = call.direction === 'in' && !call.joinedOnThisDevice;
```
adamweeks: `hasJoinedOnThisDevice` is a property that we calculate below. `joinedOnThisDevice` comes directly from the sdk's call object.

PR #847, opened 2018-10-01
fix(ciscospark): Do not display device joins/leaves in message widget 1:1

jeffhoang:


jeffhoang:
I backed out 2 previous commits and did a hard reset. Then I re-committed my changes.

PR #850, opened 2018-10-08
Journey Test Refactors

adamweeks:
Primarily using the newly creating test users helper functions to reduce the amount of code in our journey test setup.


test/journeys/lib/test-users.js
```diff
@@ -39,3 +43,139 @@ export default function setupTestUserJwt({displayName}) {
...
     });
 }
+
+/**
+ * Creates a number of test users
+ * @param {number} count number of users to generate
+ * @param {Array} config with config to pass to test user creator
+ * @returns {Array}
+ */
+export function createTestUsers(count, config) {
```
moriahmaney: very nice! thanks for making these into helper functions.

PR #864, opened 2018-11-06
External send message

adamweeks:
Builds on #862 and #863

This adds a `sendMessage` prop to the space widget that will allow you to pass a string to have it automatically send.


adamweeks:
Closes #863

PR #865, opened 2018-11-13
feat(widget-demo): add sticky mode support

adamweeks:
This will make the loaded widgets "sticky" to the bottom of the screen vs having a dedicated space at the bottom of the page.


adamweeks:
With the sticky mode enabled:
image

packages/node_modules/@ciscospark/widget-demo/src/components/demo-widget/index.js
```diff
@@ -163,6 +164,13 @@ class DemoWidget extends Component {
     return this.setState({initialActivity: value});
   }
 
+  @autobind
+  handleToggleStickyMode() {
```
adamweeks: @mattnorris since we have access to state here, we are literally toggling the mode boolean. I can change it out if you'd like.

PR #872, opened 2018-11-26
feat(r-c-button-controls): use collab-ui ActivityButtons

moriahmaney:


packages/node_modules/@ciscospark/react-component-button-controls/src/index.js
```diff
@@ -25,19 +25,31 @@ function ButtonControls(props) {
...
 
-  const buttonsRendered = buttons.map((button, index) => {
+  const buttonsRendered = buttons.map((button) => {
     const label = showLabels && button.label ? button.label : '';
-    const key = label.length ? label : `button-control-${index}`;
-    const labelPosition = showLabels ? 'bottom' : '';
+    let type;
+
+    switch (button.iconType) {
+      case 'ICON_TYPE_VIDEO_OUTLINE':
```
adamweeks: Should we still be using the constants from `@ciscospark/react-component-icon`? If we're not using our icons any more, we should probably just require the icon type from collab-ui.

PR #876, opened 2018-11-28
feat(widget-recents): Add notification level icon support

jeffhoang:


adamweeks:
@jeffhoang doing a test on the netlify deploy, everything looks good except for the blue dot behavior. @moriahmaney is there a guide for what should be "blue dotted"?

Widget:
image

Web Client:
image

packages/node_modules/@ciscospark/react-component-utils/src/badge.js
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Get the badge state of a conversation
+ * @param  {object} space - a space
+ * @param  {boolean} [unread=false] - whether this conversation has unread activities
+ * @param  {boolean} [hasMention=false] - whether this conversation has unread mentions
+ * @param  {string} [globalNotificationState=NotificationsConstants.NOTIFICATIONS_ALL]
+ *          - the value of global notification setting
+ * @returns {string} a value that indicates what the badge of this conversation should be.
+ */
+export function getBadgeState(space, unread = false, hasMention = false,
```
adamweeks: Let's convert these to an object parameter instead of multiple parameters. So `getBadgeState({space, unread = false` etc.

packages/node_modules/@ciscospark/react-component-utils/src/badge.test.js
```diff
@@ -0,0 +1,337 @@
...
+    const convo = {isOneOnOne: true};
+    expect(getBadgeState(convo, true, false)).toEqual(NotificationsConstants.NOTIFICATIONS_BADGE_UNREAD);
+  });
+
+  it('returns NOTIFICATIONS_BADGE_NONE for a read 1on1 convo', () => {
+    const convo = {isOneOnOne: true};
+    expect(getBadgeState(convo, false, false)).toEqual(NotificationsConstants.NOTIFICATIONS_BADGE_NONE);
+  });
+
+  it('returns UNREAD for an unread group convo (golobal ALL)', () => {
```
adamweeks: Spelling: find and replace all `golobal` in this file

packages/node_modules/@ciscospark/react-component-utils/src/badge.test.js
```diff
@@ -0,0 +1,337 @@
...
+    const convo = {
```
adamweeks: Nice to have: you could do a `beforeEach` and set the convo object to a default state and just do one or two lines in each test block to change the state. Would reduce the large code blocks in each block.

packages/node_modules/@ciscospark/react-component-utils/src/mentions.js
```diff
@@ -0,0 +1,20 @@
+export function hasMentions(currentUser, space) {
```
adamweeks: Add JSDoc for this please

packages/node_modules/@ciscospark/react-component-utils/src/mentions.test.js
```diff
@@ -0,0 +1,36 @@
+import {hasMentions} from '@ciscospark/react-component-utils';
+
+describe('Get has mentions', () => {
+  it('return true for @single person mention', () => {
```
adamweeks: Add a test case for `false`

packages/node_modules/@ciscospark/react-component-utils/src/constants.js
```diff
@@ -0,0 +1,13 @@
+export const NOTIFICATIONS_GLOBAL = 'NOTIFICATIONS_GLOBAL';
```
adamweeks: Thanks for adding these constants! Very helpful!

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -313,14 +320,14 @@ export class RecentsWidget extends Component {
       case 'share':
       case 'post': {
         const currentUserEmail = currentUser.email;
-        const otherUserId = space.participants
+        const otherUserId = !!space.participants && typeof space.participants.find === 'function' && space.participants
```
adamweeks: Nice catch!

packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -15,6 +15,10 @@ const getMedia = (state) => state.media;
 const getCallsById = (state) => state.media.byId;
 const getMercuryStatus = (state) => state.mercury.get('status');
 
+const MENTION_NOTIFICATIONS_ON = 'MENTION_NOTIFICATIONS_ON';
```
adamweeks: Do we want to move these to a constants file? I see it used a couple of places.

packages/node_modules/@webex/react-component-space-item/src/index.js
```diff
@@ -4,11 +4,13 @@ import PropTypes from 'prop-types';
 import {Avatar, Button, SpaceListItem} from '@collab-ui/react';
 import Timer from '@ciscospark/react-component-timer';
 import {getTeamColor} from '@ciscospark/react-component-utils';
+import * as NotificationsConstants from '@ciscospark/react-component-utils/src/constants';
```
adamweeks: You shouldn't need the `/src/constants`. If you want a group of constants, export the constants as a group from `react-component-utils/index.js`

test/journeys/lib/test-users.js
```diff
@@ -179,3 +179,113 @@ export function disconnectDevices(users) {
...
   return browser.call(() => Promise.all(promises));
 }
+
+/**
+ * Unmute message notification
+ * @param {Object} spark
+ * @param {String} convoId
+ * @returns {Object}
+ */
+export function unmuteMessageNotification(spark, convoId) {
```
adamweeks: Awesome helpers, but can we move them out of the `test-users` helper library? Maybe a `convo.js` file?

jeffhoang:
> @jeffhoang doing a test on the netlify deploy, everything looks good except for the blue dot behavior. @moriahmaney is there a guide for what should be "blue dotted"?
> 
> Widget:
> image
> 
> Web Client:
> image

Those blue dots are existing behavior for "unread" convo activities within a group or one-on-one...since we don't have a feature to set notification settings yet, the widget is set as global notification to ALL (on).

adamweeks:
@jeffhoang aren't the notifications set per-space on the server side? I never set the notifications on that space in the web client, so a new message in the space should only be bold (at least I thought so)?

jeffhoang:
This update now has logic to fetch notification related features for a given user.  This allows correct retrieval of particular global notification state, rather than just NOTIFICATION_ALL.  Note: additional journey tests to cover what is tested in badge.test.js will be a separate ticket.

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -292,6 +310,9 @@ export class RecentsWidget extends Component {
       props.fetchSpace(sparkInstance, spaceId)
         .then((newSpace) => {
           if (newSpace) {
+            // Fetch latest notification features
+            props.getFeature('user', 'group-message-notifications', sparkInstance);
```
adamweeks: You should use your constants that you created

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -193,6 +200,10 @@ export class RecentsWidget extends Component {
               });
             });
         }
+
+        // Fetch notification features
+        props.getFeature(USER, GROUP_MESSAGE_NOTIFICATIONS, sparkInstance);
```
adamweeks: This is going to fetch the features every time the props update, do we need to create a widget status for tracking?

packages/node_modules/@ciscospark/widget-recents/src/container.js
adamweeks: Do we have to fetch the user's features every time a new space loads?

packages/node_modules/@ciscospark/react-component-utils/src/constants.js
```diff
@@ -0,0 +1,22 @@
...
+export const EMAIL_NOTIFICATIONS_ONE_TO_ONES = 'EMAIL_NOTIFICATIONS_ONE_TO_ONES';
+export const EMAIL_NOTIFICATIONS_DIGEST_EMAILS = 'EMAIL_NOTIFICATIONS_DIGEST_EMAILS';
+
+export const NOTIFICATIONS_BADGE_NONE = 'NOTIFICATIONS_BADGE_NONE';
+export const NOTIFICATIONS_BADGE_UNREAD = 'NOTIFICATIONS_BADGE_UNREAD';
+export const NOTIFICATIONS_BADGE_MENTION = 'NOTIFICATIONS_BADGE_MENTION';
+export const NOTIFICATIONS_BADGE_MUTE = 'NOTIFICATIONS_BADGE_MUTE';
+export const NOTIFICATIONS_BADGE_UNMUTE = 'NOTIFICATIONS_BADGE_UNMUTE';
+
+export const USER = 'user';
```
adamweeks: Can you prefix these with `FEATURES_` so we know what they are meant for.

packages/node_modules/@ciscospark/redux-module-spaces/src/helpers.js
```diff
@@ -1,5 +1,9 @@
 const TAG_ONE_ON_ONE = 'ONE_ON_ONE';
 const TAG_HIDDEN = 'HIDDEN';
+const MENTION_NOTIFICATIONS_ON = 'MENTION_NOTIFICATIONS_ON';
```
adamweeks: Let's use the constants you setup in `@ciscospark/react-component-utils`

packages/node_modules/@webex/react-component-space-item/src/index.js
```diff
@@ -3,12 +3,21 @@ import PropTypes from 'prop-types';
...
-import {getTeamColor} from '@ciscospark/react-component-utils';
+import {
+  getTeamColor,
+  NOTIFICATIONS_BADGE_MENTION,
+  NOTIFICATIONS_BADGE_MUTE,
+  NOTIFICATIONS_BADGE_NONE,
+  NOTIFICATIONS_BADGE_UNMUTE,
+  NOTIFICATIONS_BADGE_UNREAD
+} from '@ciscospark/react-component-utils';
+// import * as NotificationsConstants from '@ciscospark/react-component-utils/src/constants';
```
adamweeks: remove the commented code

packages/node_modules/@ciscospark/widget-recents/src/container.js
jeffhoang: This is the only mercury event to my knowledge that it could use  to refresh a user's global notification (feature) updates.

packages/node_modules/@ciscospark/widget-recents/src/container.js
jeffhoang: I'll have to dig around for a widget status that it may use to only fetch once during an init stage

packages/node_modules/@ciscospark/widget-recents/src/container.js
adamweeks: Ok gotcha, could you update the code comment with that info?

packages/node_modules/@ciscospark/widget-recents/src/container.js
adamweeks: Take a look at the `isFetchingTeams` and `hasFetchedTeams` statuses. You'll want to use that same method but add a `hasFetchedNotificationsFeature`.

packages/node_modules/@ciscospark/widget-recents/src/container.js
jeffhoang: I found another approach that's somewhat more elegant.

jeffhoang:
Ok, I've split this PR into separate commits, according to scope.

adamweeks:
img

PR #878, opened 2018-11-30
Update file previews to new IA

moriahmaney:


packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
```diff
@@ -12,6 +12,7 @@ const SHARE_VERB = 'share';
 
 
 const propTypes = {
+  actor: PropTypes.object.isRequired,
```
adamweeks: Total nitpick: would be great if this was a shape instead of an object so we know what properties the object needs.

PR #893, opened 2018-12-13
feat(widget-space): add support for menu toggle

moriahmaney:


packages/node_modules/@ciscospark/widget-space/src/container.js
```diff
@@ -36,6 +35,7 @@ const injectedPropTypes = {
 
 export const ownPropTypes = {
   customActivityTypes: PropTypes.object,
+  currentActivity: PropTypes.string,
```
adamweeks: This should be in injected props since it is coming from the selector.

packages/node_modules/@ciscospark/widget-roster/src/container.js
```diff
@@ -210,6 +211,7 @@ export class RosterWidget extends Component {
     return (
       <div className={classNames('ciscospark-roster', styles.roster)}>
         <ActivityMenuHeader
+          activityTypes={activityTypes}
```
adamweeks: Can you add this to the injected prop types?

PR #894, opened 2018-12-13
feat(widget-space): add ability to toggle composer buttons

moriahmaney:


packages/node_modules/@ciscospark/react-container-message-composer/src/components/ComposerButtons.test.js
```diff
@@ -10,6 +10,9 @@ describe('ComposerButtons component', () => {
 
   beforeEach(() => {
     props = {
+      composerActions: {
```
adamweeks: can you add another test with attachFiles being false?

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
```diff
@@ -227,6 +227,7 @@ export class MessageComposer extends Component {
     const {props} = this;
     const {
       activity,
+      composerActions,
```
adamweeks: Can we add to proptypes?

packages/node_modules/@ciscospark/react-container-message-composer/src/components/ComposerButtons.test.js
moriahmaney: yeah i will. good catch, thanks!!

packages/node_modules/@ciscospark/react-container-message-composer/src/container.js
moriahmaney: eagle eye! ty for catching that :)

PR #899, opened 2018-12-18
feat(widget-recents): space type filter (hide room types)

jeffhoang:


packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -638,14 +676,24 @@ export class RecentsWidget extends Component {
       });
 
       let filteredSpacesListWithActivityText;
+      // Filter space list based on type
+      const groupBy = spaceTypeFilter === SPACE_TYPE_GROUP || spaceTypeFilter === SPACE_TYPE_ONE_ON_ONE;
```
adamweeks: Doing this all in the render function is going to take a TON of compute power for every render. I think it might be best to move this into the `getRecentSpaces` function in `selector.js`. That way our results are memoized and we don't have to compute them every time the widget renders.

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -638,14 +676,24 @@ export class RecentsWidget extends Component {
...
+      if (groupBy) {
+        filteredSpacesListWithActivityText = this.constructor.filterSpace({
```
adamweeks: FYI, to call the static function, use `RecentsWidget.filterSpace` instead.

test/journeys/specs/recents/dataApi/space-type-filter.js
```diff
@@ -0,0 +1,186 @@
+import {assert} from 'chai';
+
+import {createSpace, disconnectDevices, registerDevices, setupGroupTestUsers} from '../../../lib/test-users';
+
+import {jobNames, renameJob, updateJobStatus} from '../../../lib/test-helpers';
+import {elements} from '../../../lib/test-helpers/recents-widget';
+
+describe('Widget Recents with Space Type Filter: Data API', () => {
```
adamweeks: Can we combine all of the filter tests into one suite? No need to setup all the test users and spaces again. Need this for smoke tests and global as well.

test/journeys/specs/smoke/widget-recents/space-type-filter.js
```diff
@@ -0,0 +1,190 @@
...
+
+import {createSpace, disconnectDevices, registerDevices, setupGroupTestUsers} from '../../../lib/test-users';
+
+import {jobNames, renameJob, updateJobStatus} from '../../../lib/test-helpers';
+import {elements} from '../../../lib/test-helpers/recents-widget';
+
+describe('Widget Recents with Space Type Filter', () => {
+  const browserLocal = browser.select('browserLocal');
+  const TIMEOUT = 12000;
+  const RETRY_TOTAL = 3;
```
adamweeks: I like the idea of retries, but can we make it configurable?

packages/node_modules/@ciscospark/widget-demo/src/components/demo-widget/index.js
```diff
@@ -393,25 +407,52 @@ class DemoWidget extends Component {
...
-                >
-                  Remove Recents Widget
-                </Button>
-              </div>
+              <h3>Space Type Filter</h3>
+              <RadioGroup
+                ariaLabel="Space Type Filter"
+                name="spaceTypeFilter"
+                onChange={(e) => this.handleSpaceTypeFilterChange(e)}
+                values={[SPACE_TYPE_ALL]}
```
adamweeks: this should be `values={[this.state.spaceTypeFilter]}` so it properly maps to the state value.

test/journeys/specs/recents/dataApi/space-type-filter.js
moriahmaney: +1 on this, would save a lot of time in the tests

test/journeys/specs/smoke/widget-recents/space-type-filter.js
jeffhoang: It is configurable in wdio.conf.js under mocha options:

screen shot 2018-12-19 at 11 09 32 am

It will apply globally though and in order for your test block to work, it has to be normal function and not a fat arrow one.

An alternative is setting retries within a suite and/or individual tests: http://webdriver.io/guide/testrunner/retry.html ...

packages/node_modules/@ciscospark/widget-recents/src/container.js
jeffhoang: I've moved all filter-related functions into selector

wdio.conf.js
```diff
@@ -249,7 +248,8 @@ exports.config = {
   mochaOpts: {
     ui: 'bdd',
     timeout: mochaTimeout,
-    bail: 1
+    bail: 1,
+    retries: 3
```
adamweeks: Nice!

test/journeys/specs/recents/dataApi/space-type-filter.js
```diff
@@ -0,0 +1,186 @@
...
+  const SPACE1 = 'Test Group Space';
+  const SPACE2 = 'Ask Group Space';
+  const SPACE3 = 'General Group Space';
+  const SPACE4 = 'Ask Widgets';
+
+  let allPassed = true;
+  let docbrown, lorraine, marty, participants;
+
+  before('start new sauce session', () => {
+    renameJob(jobNames.recentsDataApi, browser);
```
adamweeks: Sorry I missed this on the previous filter PR, we'll need to create a new job name for this test so they show up on sauce properly. Just something like `jobNames.recentsFilterDataApi` added to the test-helpers file.

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -631,24 +636,22 @@ export class RecentsWidget extends Component {
       hasFetchedSpaces &&
       hasFetchedGroupMessageNotificationFeature &&
       hasFetchedMentionNotificationFeature) {
-      const spacesListWithActivityText = spacesListArray.map((space) => {
+      const showFilteredSpaces = enableSpaceListFilter
+        && keywordFilteredSpacesListArray && keywordFilteredSpacesListArray.length > 0;
+      let spacesList;
+      if (!showFilteredSpaces && !spaceTypeFilter) {
```
adamweeks: Let's do this in the selector as well, that way we only have one spacesList

packages/node_modules/@ciscospark/widget-recents/src/selector.js
```diff
@@ -126,6 +131,15 @@ const getRecentSpaces = createSelector(
...
     });
+
+    // filter space list by type
+    if (spaceType && [SPACE_TYPE_ONE_ON_ONE, SPACE_TYPE_GROUP].includes(spaceType)) {
+      return recents.filter((space) => {
+        const {type} = space;
+        return spaceType === type;
+      });
+    }
+
```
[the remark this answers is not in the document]
jeffhoang: I've tried this but it's causing huge performance issues and also seeing an error of [Violation] 'input' handler took n ms...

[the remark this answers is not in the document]
jeffhoang: Found a better solution...doing keyword filtering inside getRecentSpacesWithDetail instead.

packages/node_modules/@ciscospark/widget-recents/README.md
```diff
@@ -106,6 +106,8 @@ When loading the widgets there are some configuration options you can provide:
 | Name | Data API  | Description |
 |:--|:-------|---|
 | `logLevel` | `data-log-level` | (default: `silent`) When present, widget will log debug information to console. This can be set to: `error`, `warn`, `debug`, `info`, `trace`, or `silent` |
+| `enableSpaceListFilter` | `data-enable-space-list-filter` | (default: true) This enables a search term input box for filtering the recents widget's space list.  To disable, set this to `false`. |
+| `spaceTypeFilter` | `data-space-type-filter` | (default: `all`)  This filters the list of spaces by type. This can be set to: `direct` for one-on-one conversations, `all` for everything, or `group` for group spaces. |
```
adamweeks: Thanks for adding these!

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -687,10 +678,7 @@ export class RecentsWidget extends Component {
               isLoadingMore={isFetchingSpaces}
               onCallClick={handleCallClick}
               onClick={this.handleSpaceClick}
-              spaces={
-                enableSpaceListFilter && keywordFilter && keywordFilter.length > 0
-                  ? filteredSpacesListWithActivityText : spacesListWithActivityText
-              }
+              spaces={spacesListWithActivityText}
```
adamweeks: img

test/journeys/specs/recents/dataApi/space-type-filter.js
```diff
@@ -0,0 +1,186 @@
+import {assert} from 'chai';
```
adamweeks: Aren't these tests in the space-list-filter file now?

jeffhoang: forgot to remove this file...

PR #910, opened 2019-01-11
fix(widget-recents): padding for search input box for lower resolutions

jeffhoang:


packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -684,7 +684,7 @@ export class RecentsWidget extends Component {
                   name="pillSearchInput"
                   htmlId="pillSearchInput"
                   type="pill"
-                  inputSize="large-10"
+                  inputSize="small-10 medium-10 large-10"
```
adamweeks: This seems counter-intuitive. I thought the prop was for a single size. Was this some sort of workaround provided by collab-ui?

adamweeks: It looks like the answer you got in the Q&A space says we just need `small-10`, is that correct?

jeffhoang: Yup, this is correct...

PR #916, opened 2019-01-23
feat(recents): add extendedLoad and spaceLoadCount properties

adamweeks:


akoushke:
There are errors displayed in the console. Could be one of the reasons the app crashes

adamweeks:
@akoushke avatars are loading fine today, I think it was related to cloud drive outage

PR #919, opened 2019-02-04
fix(widget-recents): fetch proper to user avatar in direct spaces

adamweeks:
We were trying to get the "toPersonId" space detail to fetch an avatar. Unfortunately, this is a calculated property, which isn't available on the direct object, so we have to calculate it manually.

### DIFF
--- a/packages/node_modules/@ciscospark/widget-recents/src/enhancers/setup.js
+++ b/packages/node_modules/@ciscospark/widget-recents/src/enhancers/setup.js
@@ -16,20 +16,31 @@ import {
 import {updateSpaceTypeFilter, updateWidgetStatus} from '../actions';
 import getRecentsWidgetProps from '../selector';
 
-function getSpaceAvatar(s, props) {
-  const {sparkInstance} = props;
+/**
+ * Gets the avatar for a space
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/widget-recents/src/enhancers/setup.js
```diff
@@ -16,15 +16,19 @@ import {
...
 
   if (!s.isDecrypting) {
     if (
       s.type === 'direct'
-      && s.toPersonId
     ) {
-      props.fetchAvatar({userId: s.toPersonId}, sparkInstance);
+      let {toPersonId} = s;
+
+      if (!toPersonId) {
```
akoushke: do you know why this is though? why do we have to use `currectUserId`? Why is `toPersonId` undefined?

akoushke: I guess I did. okay. Would it be possible to add comments on why we are doing this?

adamweeks: @akoushke updated, thanks for the feedback.

PR #920, opened 2019-02-06
Fix recents widget not showing notification icons

adamweeks:
This was way more than a fix, sadly. It was a new implementation to handle "TAG" and "UNTAG" events that come across via websocket when someone modifies a space's notification settings.

I also moved the activity listener out of the recents container and into an enhancer.

This one is definitely easier to read commit-by-commit.


akoushke:
excuse me but I'm not following what the problem was :/

PR #923, opened 2019-02-11
fix(circleci): pin firefox version to 64

moriahmaney:


akoushke:
so we still don't know why we version 65 does not work huh?

PR #925, opened 2019-02-14
Add options to #getSpaces for storing users and activities

adamweeks:
Users and activities are not needed for rendering the recents widget. Since we fetch so many spaces, optimize the processing of the spaces and reduce the amount of writes to the redux store.

The space widget still needs these, so I made the storing configurable.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
+++ b/packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
@@ -115,9 +115,19 @@ function decryptSpace(space) {
  * @export
  * @param {Object} sparkInstance
  * @param {Object} options
+ * @param {boolean} options.deferDecrypt defer the decryption until after download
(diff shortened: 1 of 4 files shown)


packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
```diff
@@ -149,8 +158,13 @@ export function getSpaces(sparkInstance, options = {}) {
                 const participants = decryptedSpace.participants.items;
                 const activityActors = activities.map((a) => a.actor);
 
-                dispatch(storeUsers([...participants, ...activityActors]));
-                dispatch(storeActivities(decryptedActivities));
+                if (shouldStoreUsers) {
```
akoushke: I'm excited for this...... but I'm thinking if anybody would even ask for that? like what is the point of storing activities and users?

PR #927, opened 2019-02-15
Remove storing of activities and users in fetch spaces

adamweeks:
When fetching a group of spaces, we only need to store the user data from the "to user" in a one-on-one conversation. The recents widget does not need any other data in the redux store, ie: other participants and space activities. 

I've also split out `fetchSpacesEncrypted` from `fetchSpaces` to make things more readable.

### DIFF
--- a/packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
+++ b/packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
@@ -1,6 +1,5 @@
 import {storeUsers} from '@ciscospark/redux-module-users';
 import {storeActivities} from '@ciscospark/redux-module-activities';
-import {uniqBy} from 'lodash';
 
 import {
   TAG_LOCKED,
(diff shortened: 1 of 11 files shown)


akoushke:
Adam this made rendering much much faster on chrome. I had a question though out of curiosity, how did you find out we were storing unnecessary data and what changes that you made here reflect that?

packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
```diff
@@ -108,78 +117,6 @@ function decryptSpace(space) {
...
-        if (deferDecrypt && space.decrypt) {
-          decryptPromise = decryptSpace(space)
-            .then((decryptedSpace) => {
-              if (decryptedSpace) {
-                const s = Object.assign({}, decryptedSpace, {isDecrypting: false});
-                const decryptedActivities = decryptedSpace.activities.items;
-                const participants = decryptedSpace.participants.items;
-                const activityActors = activities.map((a) => a.actor);
-
-                dispatch(storeUsers([...participants, ...activityActors]));
```
adamweeks: @akoushke this is where the old code was storing users and activities. When you had the console open you would see a flood of the actions for store user and store activity come across as spaces decoded. Realized this was extra processing that wasn't needed for the recents widget at all with the new design.

akoushke: got it. thanks!

PR #932, opened 2019-02-22
fix(r-m-users): add support for users with no nicknames (bots)

adamweeks:


packages/node_modules/@ciscospark/redux-module-users/src/helpers.test.js
```diff
@@ -37,6 +37,17 @@ describe('Redux module users helpers', () => {
       expect(constructUserFromHydra(user)).toMatchSnapshot();
     });
 
+    it('converts a hydra object to user shape without a nickname', () => {
```
akoushke: Noice!!!!

PR #934, opened 2019-03-05
Add Federation support

mattnorris:
Add Federation support via the `FEDERATION` environment variable. 

[Manual testing instructions](https://confluence-eng-gpk2.cisco.com/conf/display/WBXT/Federation)


packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
```diff
@@ -71,8 +71,13 @@ function defaultConfig(options = {}) {
     },
     credentials: {
       client_id: process.env.CISCOSPARK_CLIENT_ID,
+      federation: true,
```
adamweeks: Are we expected to have federation enabled for all users of the widgets?

packages/node_modules/@ciscospark/react-redux-spark/src/spark.js
```diff
@@ -71,8 +71,13 @@ function defaultConfig(options = {}) {
...
       scope: 'spark:all spark:kms'
     },
+
+    // Federation EU cluster
+    orgId: process.env.EU_PRIMARY_ORG_ID,
```
adamweeks: I'm a bit confused by this env name. Are federation org ids all going to be "EU"?

package.json
```diff
@@ -57,7 +57,9 @@
   "homepage": "https://github.com/webex/react-ciscospark#readme",
   "dependencies": {
     "@ciscospark/common": "1.49.3",
+    "@ciscospark/common-timers": "1.49.3",
```
adamweeks: Where are these used? Any idea why we need to specifically import them now?

package.json
```diff
@@ -57,7 +57,9 @@
...
     "@ciscospark/helper-html": "1.49.3",
+    "@ciscospark/http-core": "1.49.3",
```
adamweeks: Could these work in devDependencies instead?

package.json
```diff
@@ -45,10 +45,11 @@
...
   },
   "contributors": [
-    "Adam Weeks <adweeks@cisco.com>",
-    "Bernie Zang <nzang@cisco.com>",
+    "Arash Koushkebaghi <akoushke@cisco.com>",
     "Moriah Maney <momaney@cisco.com>",
-    "Arash Koushkebaghi <akoushke@cisco.com>"
+    "Matt Norris <matnorri@cisco.com> (http://mattnorris.me)",
+    "Adam Weeks <adweeks@cisco.com>",
+    "Bernie Zang <nzang@cisco.com>"
```
adamweeks: Is there a method to your sort order madness? 😆

mattnorris: Order of awesomeness 🎸

package.json
mattnorris: There were several tests failing if I didn't include them. Specifically one of the files was the built dist of `spark-core`. It complained it couldn't find the module `common-timers` for example.

package.json
mattnorris: I don't think so because tests were failing.

package.json
adamweeks: Good call!

adamweeks:
@mattnorris could you squash 0a0cff7 and ec326f0? That way we can have a cleaner changelog.

I'm assuming this isn't ready to go to master yet?

mattnorris:
@adamweeks squashed. 

Is there a spot where widgets are hosted and point to Integration?

adamweeks:
@mattnorris removing my approval since we discussed you were going to change this to be an env var controlled feature. Will also need a rebase on master due to jenkinsfile updates.

package.json
```diff
@@ -45,36 +45,40 @@
...
-    "@ciscospark/internal-plugin-presence": "1.47.3",
-    "@ciscospark/internal-plugin-search": "1.47.3",
-    "@ciscospark/internal-plugin-team": "1.47.3",
-    "@ciscospark/internal-plugin-user": "1.47.3",
-    "@ciscospark/plugin-authorization": "1.47.3",
-    "@ciscospark/plugin-logger": "1.47.3",
-    "@ciscospark/plugin-people": "1.47.3",
-    "@ciscospark/plugin-phone": "1.47.3",
-    "@ciscospark/spark-core": "1.47.3",
+    "@ciscospark/common": "1.49.3",
```
mattnorris: @adamweeks should I bump this to the latest?

adamweeks: sure!

test/journeys/specs/smoke/multiple/index.js
```diff
@@ -1,189 +0,0 @@
-import {assert} from 'chai';
```
adamweeks: Why are we deleting this file?

mattnorris: We are not... on purpose anyway.

PR #936, opened 2019-03-06
Ignore threaded replies

adamweeks:
This ignores any conversation activity that is a reply. It ignores the activities that come across via websocket as well as the activities that are loaded with the conversation.


packages/node_modules/@ciscospark/redux-module-conversation/src/helpers.js
```diff
@@ -55,7 +57,22 @@ export function normalizeActivities(activities) {
...
-  return activities.filter((a) =>
-    !(a.verb === 'update' &&
-    a.object && a.object.objectType === 'content'));
+  return activities.filter((a) => {
+    const isUpdate = a.verb === API_ACTIVITY_VERB.UPDATE;
+    const isContent = a.object && a.object.objectType === 'content';
+    const isReply = a.type === API_ACTIVITY_TYPE.REPLY;
+    let shouldInclude = true;
+
+    // !(a.verb === 'update' &&
```
akoushke: ?

packages/node_modules/@ciscospark/widget-recents/src/enhancers/listeners.js
```diff
@@ -69,6 +69,11 @@ function processActivity(activity, space, props) {
     spacesList
   } = props;
 
+  // Reply activities are not currently supported
+  if (activity.type === API_ACTIVITY_TYPE.REPLY) {
```
akoushke: is this quoting?

adamweeks: No, this is threaded replies. We probably need a feature epic for click quoting.

packages/node_modules/@ciscospark/redux-module-conversation/src/helpers.js
adamweeks: oops, old code, will remove

adamweeks:
@akoushke I've addressed your comments. Let me know if there are any others.

packages/node_modules/@ciscospark/widget-recents/src/enhancers/listeners.js
akoushke: quoting kinda works actually. they are doing a mark ups with `>`.

akoushke:
I don't know how to test this, since the feature is not released yet. can we wait on this and then test it after the feature is released?

adamweeks:
> I don't know how to test this, since the feature is not released yet. can we wait on this and then test it after the feature is released?

No, we cannot wait because once they are enabled, it will break the widget. This is the same implementation the web client is doing.

akoushke:
> > I don't know how to test this, since the feature is not released yet. can we wait on this and then test it after the feature is released?
> 
> No, we cannot wait because once they are enabled, it will break the widget. This is the same implementation the web client is doing.

where are they enabled?

PR #938, opened 2019-03-07
Fix for firefox automation problems

taymoork2:
Adding some preference flags seems to have done the trick, but firefox needed to use my webcam for some reason?

So draft for now to iron out tweaks and fix whatever I missed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,7 +130,7 @@ jobs:
           no_output_timeout: 15m
           command: |
             set -em
-            VERSION=64 BROWSER=firefox npm run test:integration
+            BROWSER=firefox npm run test:integration
       - store_test_results:
           path: reports/junit/wdio
       - store_artifacts:
--- a/wdio.conf.js
+++ b/wdio.conf.js
@@ -6,7 +6,6 @@ const dotenv = require('dotenv');
 dotenv.config();
 dotenv.config({path: '.env.default'});
 
-
 // eslint-disable-next-line prefer-destructuring
 const argv = require('yargs').argv;
 
@@ -16,6 +15,7 @@ const {inject} = require('./scripts/tests/openh264');
 const beforeSuite = require('./scripts/tests/beforeSuite');
 
 const port = process.env.PORT || 4567;
+const isSauceEnabled = (process.env.SAUCE === 'true');
 let baseUrl = process.env.JOURNEY_TEST_BASE_URL;
 
 if (!baseUrl) {
@@ -35,7 +35,7 @@ const screenResolution = platform.toLowerCase().includes('os x') || platform ===
 
 const chromeCapabilities = {
   browserName: 'chrome',
-  chromeOptions: {
+  'goog:chromeOptions': {
     args: [
       '--use-fake-device-for-media-stream',
       '--use-fake-ui-for-media-stream',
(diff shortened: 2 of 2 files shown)


wdio.conf.js
```diff
@@ -47,10 +48,30 @@ const chromeCapabilities = {
...
 };
 const firefoxCapabilities = {
-  browserName: 'firefox'
+  browserName: 'firefox',
+  'moz:firefoxOptions': {
+    args: [
+      '-start-debugger-server',
+      '9222'
+    ],
+    ...(!isSauceEnabled && {
```
adamweeks: We add sauce capabilities at the end, check line 320ish. Would be good to group these changes here.

wdio.conf.js
```diff
@@ -35,7 +36,7 @@ const screenResolution = platform.toLowerCase().includes('os x') || platform ===
 
 const chromeCapabilities = {
   browserName: 'chrome',
-  chromeOptions: {
+  'goog:chromeOptions': {
```
adamweeks: When did this change?

wdio.conf.js
```diff
@@ -47,10 +48,30 @@ const chromeCapabilities = {
   }
 };
 const firefoxCapabilities = {
-  browserName: 'firefox'
+  browserName: 'firefox',
+  'moz:firefoxOptions': {
+    args: [
+      '-start-debugger-server',
```
adamweeks: What is this needed for?

.circleci/config.yml
```diff
@@ -130,7 +130,7 @@ jobs:
           no_output_timeout: 15m
           command: |
             set -em
-            VERSION=64 BROWSER=firefox npm run test:integration
+            VERSION=65 BROWSER=firefox npm run test:integration
```
adamweeks: Can we remove the version indicator and have it use the latest? The version 64 was added due to 65 breaking.

adamweeks:
image
Woo!

wdio.conf.js
taymoork2: w3 standardized webdriver

wdio.conf.js
taymoork2: Moved down to saucelabs, this enables extendedDebugging

.circleci/config.yml
taymoork2: Was gonna ask if this should be latest, removed

wdio.conf.js
```diff
@@ -317,6 +328,19 @@ if (isSauceEnabled) {
...
       maxDuration: 3600,
       seleniumVersion: '3.4.0',
+      extendedDebugging: true,
+      // extended debugging
+      'moz:firefoxOptions': {
+        args: [
+          '-start-debugger-server',
+          '9222'
+        ],
+        prefs: {
```
adamweeks: Will this overwrite the options set in line 52?

taymoork2: Yes I double checked before pushing because I wondered the same thing
`Object.assign(Obj1, Obj2, ..., source)`  so in our case `{}` gets overwritten by <-- `config` which gets overwritten by <-- whatever we put in there

adamweeks: In that case, do we need all those options redefined here?

taymoork2: For Firefox yes
There's no need for those options locally (as they only pertain to extendedDebugging/SauceLabs)
The options defined earlier are only necessary for local tests to run and are unnecessary for SauceLabs

adamweeks: Awesome, sounds good!

adamweeks:
@taymoork2 you'll need to rebase this on the latest master to get it through the pipeline since Jenkinsfile was updated in master.

PR #943, opened 2019-03-21
Upgrade to collab-ui 12 and jest 23

adamweeks:
Jest needed to be upgraded to add support for snapshots with react fragments.


adamweeks:
Added workaround for https://github.com/collab-ui/collab-ui/issues/148

PR #945, opened 2019-03-25
chore(tab): don't install nvm on tab tests

adamweeks:
TAB tests are failing due to the old nvm install command.


Jenkinsfile.tab
```diff
@@ -22,19 +22,15 @@ pipeline {
         withCredentials([
           string(credentialsId: 'WIDGETS_NPM_TOKEN', variable: 'WIDGETS_NPM_TOKEN')
         ]) {
+          sh 'echo \'//registry.npmjs.org/:_authToken=${WIDGETS_NPM_TOKEN}\' >> .npmrc'
```
adamweeks: This install stage is taken directly from our TAP tests script.

PR #946, opened 2019-03-26
 fix(widget-recents): recents widget does not load for user with no spaces

moriahmaney:
Fix bug where recents widget would get stuck on the loading spinner for a user with no spaces + implement no spaces design.

The CSS is borrowed from the web client to make sure we're in line with their design -- if y'all want any changes just let me know.

There is currently no better way to handle the no spaces SVG as our usual icon component can only handle SVGs with one path, not two. Collab-ui doesn't currently have an icon for it, either.

### DIFF
--- a/packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.css
+++ b/packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.css
@@ -1,0 +1,29 @@
+.noSpacesWrapper {
+  height: 100%;
+  text-align: center;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  justify-content: center;
+  background-color: black;
+}
+
+.noSpacesImg {
+  height: 72px;
+  width: 100%;
+  margin-bottom: 24px;
+}
+
+.noSpacesTitle {
+  color: #fff;
+  font-weight: 700;
+  line-height: 24px;
+  text-align: center;
+  margin-bottom: 12px;
+}
+
+.noSpacesMessage {
+  color: hsla(0,0%,100%,.6);
+  line-height: 22px;
+  text-align: center;
+}
--- a/packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.js
+++ b/packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import styles from './NoSpaces.css';
+
+const propTypes = {
+  title: PropTypes.string.isRequired,
+  emptyMessage: PropTypes.string.isRequired
+};
+
+function NoSpaces(props) {
+  const {
+    title,
+    emptyMessage
+  } = props;
+
+  return (
+    <div className={classNames('ciscospark-recents-widget-no-spaces', styles.noSpacesWrapper)}>
+      <div className={classNames('ciscospark-no-spaces-img', styles.noSpacesImg)}>
+        <svg width="80px" height="62px" viewBox="0 0 80 62" version="1.1" xmlns="http://www.w3.org/2000/svg">
+          <defs />
+          <g id="Illustrations" stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+            <g id="Artboard" transform="translate(-1053.000000, -175.000000)" fill="#FFFFFF">
+              <g id="No-Space" transform="translate(1053.000000, 175.000000)">
+                <path d="M56.999,0 C44.297,0 34,10.297 34,22.999 C34,26.583 34.82,29.975 36.282,32.998 L37.822,32.998 L56.5,32.998 C56.775,32.998 57,33.223 57,33.498 C57,33.773 56.775,33.998 56.5,33.998 L38.353,33.998 L36.797,33.998 C40.697,41.147 48.281,45.998 56.999,45.998 C69.701,45.998 79.998,35.701 79.998,22.999 C79.998,10.297 69.701,0 56.999,0" id="fill" opacity="0.400000006" />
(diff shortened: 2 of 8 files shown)


packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.css
```diff
@@ -0,0 +1,30 @@
...
+
+.noSpacesImg {
+  height: 72px;
+  width: 100%;
+  margin-bottom: 24px;
+}
+
+.noSpacesTitle {
+  color: #fff;
+  font-family: CiscoSansTT Bold,Helvetica Neue,Arial;
```
adamweeks: Could we use a collab-ui `h2` to give us most of this styling (aside from the alignment and padding)?

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.css
```diff
@@ -0,0 +1,30 @@
...
+  font-size: 16px;
+  line-height: 24px;
+  text-align: center;
+  margin-bottom: 12px;
+}
+
+.noSpacesMessage {
+  color: hsla(0,0%,100%,.6);
```
adamweeks: Is this color code from the web client? Could we use a collab style instead?

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.js
```diff
@@ -0,0 +1,46 @@
...
+import classNames from 'classnames';
+
+import styles from './NoSpaces.css';
+
+const propTypes = {
+  title: PropTypes.string.isRequired,
+  emptyMessage: PropTypes.string.isRequired
+};
+
+const defaultProps = {
```
adamweeks: No need for an empty default props

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+function NoSpaces(props) {
+  const {
+    title,
+    emptyMessage
+  } = props;
+
+  return (
+    <div className={classNames('ciscospark-recents-widget', styles.noSpacesWrapper)}>
```
adamweeks: `ciscospark-recents-widget` is the class name of the main recents widget. I would expect it to be `ciscospark-recents-widget-no-spaces` perhaps.

packages/node_modules/@ciscospark/widget-recents/src/container.js
```diff
@@ -237,13 +240,15 @@ export class RecentsWidget extends Component {
...
-    const isReady = spacesListArray && (spacesListArray.length || isFiltered);
+    const isReady = (widgetStatus.hasFetchedInitialSpaces || isFiltered);
 
     const isLoadingMore = props.extendedLoad
       ? !widgetStatus.hasFetchedAllSpaces : !widgetStatus.hasFetchedInitialSpaces;
 
     // Display the header if any of these options are true, otherwise, hide header.
     const showHeader = enableSpaceListFilter || enableAddButton || enableUserProfile;
+    const hasSpaces = widgetStatus.hasFetchedInitialSpaces && spacesListArray.length > 0;
+    const emptyMessage = !hasSpaces && enableAddButton ? 'Create a space using the plus button next to the search bar above.' : 'Create spaces in Webex Teams to see them here.';
```
adamweeks: Can these messages be added to the `messages.js` for eventual internationalization?

test/journeys/lib/test-helpers/recents-widget/index.js
```diff
@@ -4,8 +4,10 @@ import waitForPromise from '../../wait-for-promise';
 
 export const elements = {
   loadingScreen: '.ciscospark-loading',
-  recentsWidget: '.ciscospark-spaces-list-wrapper',
```
adamweeks: lolwut.

test/journeys/specs/smoke/widget-recents/index.js
```diff
@@ -105,7 +103,24 @@ describe('Smoke Tests - Recents Widget', () => {
...
+    });
+
+    it('has no spaces message', () => {
+      assert.isTrue(browserLocal.element(elements.noSpacesMessage).isVisible(), 'does not have no spaces message');
+      assert.equal(browserLocal.element(elements.noSpacesMessage).getText(), 'Create a space using the plus button next to the search bar above.');
+    });
+  });
+
   describe('Group Space', () => {
+    it('creates spaces', () => {
```
adamweeks: Want to try switching back to `before` blocks or is it a hopeless cause with wdio v4 + multiremote?

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.js
moriahmaney: oops, good catch. copy pasta i forgot to delete :)

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.css
moriahmaney: color code is from web client, not sure what the matching collab style is (if there is one)

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.css
moriahmaney: will try it and get back to you

test/journeys/lib/test-helpers/recents-widget/index.js
moriahmaney: that class is only present when the spaces list has at least one space so i swapped to the `ciscospark-recnts-widget` instead

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.js
moriahmaney: yeah, that's what i had at first i think i was just testing why the recents tests weren't working, will put it back

packages/node_modules/@ciscospark/widget-recents/src/container.js
moriahmaney: yepp, will do.

test/journeys/specs/smoke/widget-recents/index.js
moriahmaney: resolved via teams to leave for now

packages/node_modules/@ciscospark/widget-recents/src/components/NoSpaces.css
moriahmaney: update: no matching collab-ui style

PR #949, opened 2019-03-27
Enable federation by default

mattnorris:
We will push this when the clients turn on federation.


mattnorris:
@adamweeks yes. Enabling federation just means that _if_ the user is fed, it sends out a request to get the proper URLs to use.

adamweeks:
@mattnorris is this still waiting?

PR #950, opened 2019-04-04
fix(journeys): fix firefox messaging journey test

moriahmaney:


test/journeys/specs/tap/widget-space/oneOnOne.js
```diff
@@ -45,6 +45,13 @@ describe('Widget Space: One on One: TAP', () => {
     remote.browser.waitForExist(`[placeholder="Send a message to ${local.displayName}"]`, 30000);
   });
 
+  before('sticky widgets', () => {
```
adamweeks: `stick widgets to bottom of viewport` would help me to understand why we are doing this

moriahmaney: good call, thanks friend

PR #958, opened 2019-04-18
Build packages for NPM using rollup

adamweeks:
The way we are currently building everything for npm is to run webpack on each package directory, which builds EVERYTHING, including dependencies. BUT, we don’t need to do that in a module system because npm can get the dependencies. This was bad because it made people download MEGs of data during an npm install, it also didn’t allow npm users of the widgets to dedupe and tree-shake existing packages. It also caused data files (images, fonts, etc) to not be able to be loaded because they weren’t in the correct path.

Using rollup for esm config, we leave the importing of modules to npm, but we still convert our code to browser supported ES5. it drops the package size to a few K, as well as reduces the build time.

I've also gone ahead and manually bumped to version `0.2.0` so anyone that may be using the existing npm version will know this is a big change.


akoushke:
is there anyway to test this?

PR #959, opened 2019-04-23
Recents Widget: add basicMode feature

adamweeks:
Features added when basic mode is enabled:

* Spaces will be fetched from the hydra /rooms endpoint
* No decryption needed

Features removed for basic mode:

* No space avatars (hydra will be adding this feature to /rooms)
* No calling support (knowing when a space has a call)
* No Read/Unread status (spaces will always display as unread)
* No notification icons (cannot know when spaces are muted)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1084,6 +1084,15 @@
         }
       }
     },
+    "@ciscospark/plugin-rooms": {
+      "version": "1.50.18",
+      "resolved": "https://registry.npmjs.org/@ciscospark/plugin-rooms/-/plugin-rooms-1.50.18.tgz",
+      "integrity": "sha512-ii8DIG244zqkyhaNTYvp4jPUI16tLTFsEctqkx9a6CsdWgJvHMRODx3ll+L+nnJr4+IgGlUFKNsfUmYPI/T+jg==",
+      "requires": {
+        "@ciscospark/spark-core": "1.50.18",
+        "envify": "^4.1.0"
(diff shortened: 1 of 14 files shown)


akoushke:
space avatars perhaps?

akoushke:
random question, do you know if we have a JIRA for hydra to add this feature?

akoushke:
Correct me if I'm wrong, but I thought in our meeting with David and Olivier, we did not cover all the details since we had to leave the room. with that being said, I understand why you don't think getting avatars for direct spaces are not worth it since we have to make nested calls (one for memberships and then one for person details). Maybe we should not do that but can we at least get the team names of the spaces if available? my idea of `basic mode` feature was  to provide as much as possible without any encryption/decryption involved. maybe `basic mode` is not a very appropriate name to pick for what we are offering. I also think that our addin buys exactly 4 seconds before displaying the list of the rooms. Although it might not be enough, but it is buying us some time. let me know what you think?

akoushke:
IMG_3311
any possibility we can get the initial of the spaces? like AW for you, to be more Desktop alike?

adamweeks:
> any possibility we can get the initial of the spaces? like AW for you, to be more Desktop alike?
> 
> For instance Adam Kasley should be AK. I think cui take care of it. we might not be passing the right props to spaces.

I investigated this, but it is due to the collab-ui's `SpaceListItem` manually setting the avatar type.

PR #963, opened 2019-05-06
fix(widget-recents): load space avatars properly

adamweeks:
Avatars and people were not being loaded properly because the `encryptedSpaces` object wasn't being passed properly.

One too many `.thens` in the code


akoushke:
is this the reason we are getting `untitled` 1:1 rooms?

adamweeks:
> is this the reason we are getting `untitled` 1:1 rooms?

@akoushke yes, sorry, it is the bug from SPARK-72054

akoushke:
I wonder why it did not happen before :-??

PR #965, opened 2019-05-09
Momentum ui update

adamweeks:
There's very little more in this PR than simply renaming imports and updating snapshots. I'll point out the changes as in-line comments.

Edit/Update: There have been more changes due to webpack loading issues. I've also added the `md` class name to the top of the widgets using the momentum module css as it is required to get the momentum styling below it.


samples/Main.js
```diff
@@ -11,6 +11,9 @@ function Main() {
       <BasicComponents />
       <h2>Recents Components</h2>
       <RecentsComponents />
+      <h2>Base Elements that should not be modified by momentum css</h2>
```
adamweeks: Added these elements to the sample page to verify they weren't getting style overrides from momentum.

packages/node_modules/@ciscospark/react-component-button-controls/src/index.js
```diff
@@ -34,25 +34,31 @@ function ButtonControls(props) {
...
-  const buttonsRendered = buttons.map((button) => {
+  const buttonsRendered = buttons.map((button, idx) => {
     const label = showLabels && button.label ? button.label : '';
+    const ariaLabel = button.accessibilityLabel || button.label;
+    const key = button.key || `button-controls-${idx}`;
 
     if (button.callControl) {
       return (
         <CallControl
-          {...button}
```
adamweeks: When we passed all the props as an object, momentum caused react errors, so I manually specified the exact props to pass.

packages/node_modules/@webex/react-component-space-item/src/momentum.scss
```diff
@@ -0,0 +1,9 @@
+$brand-font-folder: "~@momentum-ui/core/fonts";
+$icon-font-path: "~@momentum-ui/icons/fonts";
+
+@import '~@momentum-ui/core/scss/momentum-ui-components';
+
+// Workaround for missing style
+.md-decrypting {
```
adamweeks: This style is not currently included in the momentum-ui-components package. They are adding it later.

packages/node_modules/@ciscospark/widget-space/src/index.html
```diff
@@ -4,56 +4,51 @@
 <head>
   <title>Space Widget Demo</title>
   <meta charset="utf8"/>
-  <script nomodule src="https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/6.26.0/polyfill.min.js" integrity="sha256-WRc/eG3R84AverJv0zmqxAmdwQxstUpqkiE+avJ3WSo=" crossorigin="anonymous"></script>
 </head>
 <body>
 <div style="float: left;">
-  <h2>Browser Globals</h2>
+  <h2>Space Widget</h2>
```
adamweeks: This is what gets served when you run `npm run serve:package widget-space`. It is a good way to check styles outside of the widget-demo. We've had the demo add styles that the individual widgets did not get and a bug was missed in the past. I've made this page a little more user friendly and less relying on our original env vars.

packages/node_modules/@ciscospark/widget-demo/src/components/demo-widget/momentum.scss
```diff
@@ -0,0 +1,5 @@
+$brand-font-folder: "~@momentum-ui/core/fonts";
+$icon-font-path: "~@momentum-ui/icons/fonts";
+
+// importing the base momentum so we can style the demo page
```
adamweeks: We are not importing the momentum component styles here because we want the full momentum overrides for the demo.

akoushke:
holy commits! 🙏🏻

packages/node_modules/@ciscospark/react-component-activity-item-base/src/index.js
```diff
@@ -92,7 +92,7 @@ function ActivityItemBase({
       <div className={classNames(getActionClassNames({highlight: isFlagged, flagActionPending: isFlagPending}))}>
         <Icon
           ariaLabel="Flag for follow-up"
-          color="black"
+          color="black-100"
```
akoushke: what does 100 means?

packages/node_modules/@ciscospark/react-component-button-controls/src/index.js
akoushke: we should let them know about this as well I would suggest.

packages/node_modules/@ciscospark/react-component-activity-item-base/src/index.js
adamweeks: This was due to a warning message that was displayed:

> getColorValue.js:71 [@momentum-ui] Design system colors update: black is now black-100. Refer to https://momentum.design/styles/color/style

packages/node_modules/@ciscospark/react-component-button-controls/src/index.js
adamweeks: It appears as if this is [by design](https://github.com/momentum-design/momentum-ui/blob/master/react/src/lib/Button/index.js#L180) but we can let them know for sure.

packages/node_modules/@ciscospark/react-component-activity-item-base/src/index.js
adamweeks: After discussing with the Momentum team, this warning is a bug and it should still be `black`. Will revert.

akoushke:
So the color codes I'm assuming that have changed.

**BEFORE**
Screen Shot 2019-05-13 at 9 45 14 AM
**AFTER**
Screen Shot 2019-05-13 at 9 45 07 AM

Just to make sure that you are aware of this

akoushke:
Also, something not related to this PR but it be very nice to fix is the header on the top. It looks way off. would it be possible to give it margin or something?
Screen Shot 2019-05-13 at 9 45 20 AM

adamweeks:
> So the color codes I'm assuming that have changed.

Yes, the colors have been slightly modified via momentum.

adamweeks:
> Also, something not related to this PR but it be very nice to fix is the header on the top. It looks way off. would it be possible to give it margin or something?
> Screen Shot 2019-05-13 at 9 45 20 AM

This is using the `TopBar` component from momentum. You could create a github issue with the changes you'd like if you're interested.

akoushke:
> > Also, something not related to this PR but it be very nice to fix is the header on the top. It looks way off. would it be possible to give it margin or something?
> > Screen Shot 2019-05-13 at 9 45 20 AM
> 
> This is using the `TopBar` component from momentum. You could create a github issue with the changes you'd like if you're interested.

Yup, right here: https://github.com/momentum-design/momentum-ui/issues/239

akoushke:
Looking good on IE11

adamweeks:
Passing when run outside circle...

>[undefined #0-0] 83 passing (9m, 38s)

PR #966, opened 2019-05-14
feat(widget-recents): remove extendedLoad

moriahmaney:


package-lock.json
```diff
@@ -10505,8 +10505,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
```
adamweeks: Could you remove the package lock change since we didn't have any package.json changes?

PR #973, opened 2019-05-28
feat(Core): Added a check job to ensure that our prod code is ES5 com…

akoushke:
This is my understanding  from the ticket Adam. Please let me know what you think of it.


akoushke:
will def change the commit message.

Jenkinsfile
```diff
@@ -186,6 +186,12 @@ ansiColor('xterm') {
             }
           }
 
+          stage('Check for ES5 Compatibility') {
+            sh'''
+              node_modules/.bin/es-check es5 packages/node_modules/@ciscospark/**/dist/*.js
```
adamweeks: This should also check the `/esm` and `/cjs` folders. The `esm` folder is what webpack and other module loaders will load and aren't usually configured to transpile code. 

We should also add this step to the [`publish:components` helper](https://github.com/webex/react-ciscospark/blob/cd79c7d54bcd5728d9d68e5401d8af1fca67ae1c/Jenkinsfile#L246) to verify those esm and cjs folders are good

Jenkinsfile
```diff
@@ -186,6 +186,12 @@ ansiColor('xterm') {
             }
           }
 
+          stage('Check for ES5 Compatibility') {
+            sh'''
```
adamweeks: Could you add this to the end of the `Build for CDN` stage instead since it is checking the files built in that stage.

akoushke: I thought i did.

Jenkinsfile
akoushke: will need to talk more about this tomorrow.

akoushke:
> Might be helpful to have these folder checks as npm scripts as well so we can manually run them easier.

yeah that should not be a problem

akoushke:
@adamweeks take a look please.

Jenkinsfile
adamweeks: Instead of creating a new stage, this verification step should be in the `Build for CDN` stage instead.

package.json
```diff
@@ -27,6 +27,8 @@
     "static-analysis": "npm run eslint",
     "lint:eslint": "eslint --ignore-path .gitignore $(npm run --silent lint:eslint:ci-reporter)",
     "eslint": "eslint ./",
+    "es5-check:all": "es-check es5 packages/node_modules/@ciscospark/**/dist/*.js",
+    "es5-check:esm": "es-check es5 packages/node_modules/@ciscospark/**/esm/*.js",
```
adamweeks: This needs to be updated to reflect that esm files are being built in the `/es/` folder. 

Also, when run, it is failing:

>ES-Check Error:
          ----
          · erroring file: packages/node_modules/@ciscospark/widget-space/es/index.js
          · error: SyntaxError: The keyword 'import' is reserved (1:0)
          · see the printed err.stack below for context
          ----

Looks like it supports a `--module` flag to allow `import` statements, but it doesn't work with the `es5` flag... I'm ok with changing this to:

`es-check --module es6 packages/node_modules/@ciscospark/**/es/*.js`

package.json
```diff
@@ -27,6 +27,8 @@
     "static-analysis": "npm run eslint",
     "lint:eslint": "eslint --ignore-path .gitignore $(npm run --silent lint:eslint:ci-reporter)",
     "eslint": "eslint ./",
+    "es5-check:all": "es-check es5 packages/node_modules/@ciscospark/**/dist/*.js",
```
adamweeks: `es5-check:dist` seems to be more consistent with what this is actually doing.

adamweeks: We have components in the `@webex` folder which should be checked as well.

Jenkinsfile
akoushke: ohhhh.. I got you. I misread the comment AGAIN.

akoushke:
commits are not ready yet. will fix them after an approval

PR #975, opened 2019-05-29
chore(circleci): run journey tests serially to reduce timeouts

adamweeks:
image

There seems to be issues when circle runs the integration tests at the same time on different browsers. It could be purely coincidence, but this seems to have passed easier.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,4 +161,4 @@ workflows:
             - build_for_tests
       - journey_tests_firefox:
           requires:
-            - build_for_tests
+            - journey_tests_chrome


.circleci/config.yml
```diff
@@ -161,4 +161,4 @@ workflows:
             - build_for_tests
       - journey_tests_firefox:
           requires:
-            - build_for_tests
+            - journey_tests_chrome
```
akoushke: we are testing chrome engine on firefox? I don't get this.

adamweeks: This just defines the work flow. It means that it requires the journey tests for chrome to be completed before moving on to the firefox tests.

akoushke: got it

PR #976, opened 2019-05-30
feat(widget-demo): Added a validator to validate the load space count input

akoushke:


packages/node_modules/@ciscospark/widget-demo/src/components/demo-widget/index.js
```diff
@@ -225,7 +226,18 @@ class DemoWidget extends Component {
 
   @autobind
   handleRecentSpaceLoadCountChange(e) {
-    this.setState({spaceLoadCount: e.target.value ? e.target.value : ''});
+    const isEmpty = e.target.value === '';
+    const isValid = e.target.value >= 25;
+    const errors = isEmpty ? [{error: 'Value can not be empty', type: 'error'}] : [];
+
+    if (!isValid) {
+      errors.push({error: 'Value must be greater or equal than 25.', type: 'error'});
```
adamweeks: I think `Value must be greater than or equal to 25.` reads a bit better, but not a blocker.

akoushke: yes. You are right.

PR #977, opened 2019-06-04
Webex SDK Plugin check

akoushke:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-69846

### DIFF
--- a/packages/node_modules/@ciscospark/react-redux-spark/src/component.js
+++ b/packages/node_modules/@ciscospark/react-redux-spark/src/component.js
@@ -2,6 +2,8 @@ import {Component} from 'react';
 import PropTypes from 'prop-types';
 import {bindActionCreators} from 'redux';
 import {connect} from 'react-redux';
+import {addError} from '@ciscospark/redux-module-errors';
+
 
 import {
   updateSparkStatus,
@@ -13,7 +15,8 @@ import createSpark, {createSparkJwt} from './spark';
 const injectedPropTypes = {
   spark: PropTypes.object.isRequired,
   storeSparkInstance: PropTypes.func.isRequired,
-  updateSparkStatus: PropTypes.func.isRequired
+  updateSparkStatus: PropTypes.func.isRequired,
+  addError: PropTypes.func.isRequired
 };
 
 const propTypes = {
@@ -27,11 +30,14 @@ const defaultProps = {
   guestToken: ''
 };
 
-class SparkComponent extends Component {
+const PLUGINS = ['authorization', 'logger', 'people', 'phone', 'rooms'];
+const INTERNAL_PLUGINS = ['conversation', 'feature', 'flag', 'mercury', 'presence', 'search', 'team'];
+
+export class SparkComponent extends Component {
   constructor(props) {
     super(props);
 
-    this.storeAndListen = this.storeAndListen.bind(this);
+    this.loadSparkInstance = this.loadSparkInstance.bind(this);
   }
 
   componentDidMount() {
@@ -45,10 +51,10 @@ class SparkComponent extends Component {
 
     if (!sparkInstance) {
       if (guestToken) {
-        createSparkJwt(guestToken, this.props).then(this.storeAndListen);
+        createSparkJwt(guestToken, this.props).then(this.loadSparkInstance);
       }
       else {
-        createSpark(accessToken, this.props).then(this.storeAndListen);
+        createSpark(accessToken, this.props).then(this.loadSparkInstance);
       }
     }
     else {
@@ -56,7 +62,12 @@ class SparkComponent extends Component {
     }
   }
 
-  setupDevice(spark) {
+  /**
+   * Register the device if the user has been authenticated
+   * and the device is not regusterd yet.
+   * @param {Object} sparkInstance
+   */
+  setupDevice(sparkInstance) {
     const {
       authenticated,
       registered,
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
```diff
@@ -7,14 +7,44 @@ import {createMockStore} from '@ciscospark/react-redux-spark-fixtures';
 import SparkComponent from './component';
 
 describe('spark component', () => {
-  it('renders correctly', () => {
-    const store = createMockStore();
-    const component = renderer.create(
+  let sparkInstance, store, component;
+
+  beforeEach(() => {
```
akoushke: In order for me to test the components methods, I need to use the `enzyme` library. where its not used in this project. unless you guys know other ways to test the methods without enzyme.

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
```diff
@@ -7,14 +7,44 @@ import {createMockStore} from '@ciscospark/react-redux-spark-fixtures';
...
+        feature: 'feature',
+        flag: 'flag',
+        mercury: 'mercury',
+        presence: 'presence',
+        search: 'search',
+        team: 'team'
+      }
+    };
+    component = renderer.create(
       <Provider store={store}>
```
akoushke: I definitely prefer to not use the `Provider` here. any ideas to make this work with mock `store`?

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
adamweeks: I think if you export the unconnected class you should be able to test a `new` instance of it without using enzyme. I'm not opposed to adding enzyme tests though.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -28,73 +28,68 @@ const defaultProps = {
...
-  }
-
-  shouldComponentUpdate(nextProps) {
-    const {props} = this;
-
-    return nextProps.spark !== props.spark;
-  }
-
-  static setupDevice(spark, props) {
+  setupDevice(spark) {
```
adamweeks: Would be great to add some jsdoc to these methods so we won't be as confused in the future as we were this week 🤣

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -86,10 +92,37 @@ class SparkComponent extends Component {
...
 
   storeAndListen(sparkInstance) {
+    this.checkPlugins(sparkInstance);
     this.props.storeSparkInstance(sparkInstance);
     this.listenToSparkEvents(sparkInstance);
   }
 
+  checkPlugins(instance) {
+    for (const plugin of PLUGINS) {
+      // eslint-disable-next-line no-prototype-builtins
```
adamweeks: Let's follow the suggestion from the [eslint error description](https://eslint.org/docs/rules/no-prototype-builtins) and use `Object.prototype.hasOwnProperty.call(instance, plugin);`.

Also, we've tried to prefix eslint disables with `// eslint-disable-reason my reason for disabling` in the code base.

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
akoushke: that way to mock the spark instance, which is huge. I would like to add to our new testing widget design epic.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -86,10 +92,37 @@ class SparkComponent extends Component {
   }
 
   storeAndListen(sparkInstance) {
+    this.checkPlugins(sparkInstance);
     this.props.storeSparkInstance(sparkInstance);
     this.listenToSparkEvents(sparkInstance);
   }
 
+  checkPlugins(instance) {
+    for (const plugin of PLUGINS) {
```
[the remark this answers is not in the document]
akoushke: yeah, I was going to do that but `this` scope changes and i have to add more code.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
akoushke: Screen Shot 2019-06-05 at 9 16 38 AM

adamweeks: Please add this as an `// eslint-disable-reason` comment if you are not going to use the eslint preferred method.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -27,71 +30,116 @@ const defaultProps = {
...
+          id: 'webex-plugins-missing',
+          displayTitle: 'Something\'s not right. Please try again',
+          displaySubtitle: `Webex SDK instance plugin ${plugin} is not injected to the widget properly`,
+          temporary: false
+        });
+      }
+    }
+
+    for (const internalPlugin of INTERNAL_PLUGINS) {
+      // eslint-disable-next-line no-prototype-builtins
```
adamweeks: This needs a `// eslint-disable-reason` if you are not going to use the eslint preferred method.

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
```diff
@@ -4,17 +4,83 @@ import {Provider} from 'react-redux';
...
+    });
+  });
+
+  it('verifyPlugins() verifies all external plugins to be injected properly to webex SDK instance', () => {
+    sparkInstance.verifyPlugins(spark);
+    expect(sparkInstance.props.addError).not.toHaveBeenCalled();
+  });
+
+  it('verifyPlugins() verifies all internal plugins to be injected properly to webex SDK instance', () => {
+    sparkInstance.verifyPlugins(spark);
```
adamweeks: This is the same test as the one above it.

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
```diff
@@ -4,17 +4,83 @@ import {Provider} from 'react-redux';
...
+    expect(sparkInstance.props.addError).not.toHaveBeenCalled();
+  });
+
+  it('verifyPlugins() throws an error if one of the external plugin is not injected properly', () => {
+    delete spark.logger;
+    sparkInstance.verifyPlugins(spark);
+    expect(sparkInstance.props.addError).toHaveBeenCalled();
+  });
+
+  it('verifyPlugins() throws an error if one of the external plugin is not injected properly', () => {
```
adamweeks: This is a duplicate test

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
```diff
@@ -4,17 +4,83 @@ import {Provider} from 'react-redux';
...
+        conversation: 'conversation',
+        feature: 'feature',
+        flag: 'flag',
+        mercury: 'mercury',
+        presence: 'presence',
+        search: 'search',
+        team: 'team'
+      }
+    };
+    component = renderer.create(
```
adamweeks: No need to do this in the before each since we are only using it in one test.

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
```diff
@@ -4,17 +4,83 @@ import {Provider} from 'react-redux';
...
-        <SparkComponent accessToken="abc123" />
+        <SparkConnectedComponent
+          spark={spark}
+          storeSparkInstance={jest.fn()}
+          updateSparkStatus={jest.fn()}
+          addError={jest.fn()}
+        />
       </Provider>
     );
+    sparkInstance = new SparkComponent({
```
adamweeks: Nice solution!

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -27,71 +30,116 @@ const defaultProps = {
...
-    const {props} = this;
     const {
       accessToken,
-      guestToken
+      guestToken,
+      spark
     } = this.props;
 
-    const spark = props.spark.get('spark');
+    const sparkInstance = spark.get('spark');
```
adamweeks: I like calling it `sparkInstance` to differ from the spark redux store. Good idea!

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -27,71 +30,116 @@ const defaultProps = {
...
-    const {props} = this;
-
-    return nextProps.spark !== props.spark;
-  }
-
-  static setupDevice(spark, props) {
+  /**
+   * Register the device if the user has been authenticated
+   * and the device is not regusterd yet.
+   * @param {Object} spark
```
adamweeks: Let's update this to be `sparkInstance` too so we have consistency throughout the file.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -27,71 +30,116 @@ const defaultProps = {
...
+      this.props.registerDevice(spark);
+    }
+  }
+
+  /**
+   * Listen to any new events and update the sprak instance
+   * status accordingly
+   * @param {Object} spark
+   */
+  listenToSparkEvents(spark) {
```
adamweeks: Let's update this to be `sparkInstance` too so we have consistency throughout the file.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
```diff
@@ -27,71 +30,116 @@ const defaultProps = {
...
+    this.props.storeSparkInstance(sparkInstance);
+    this.listenToSparkEvents(sparkInstance);
+  }
+
+  /**
+   * Verify if all the plugins have been injected
+   * properly to the webex SDK instance
+   * @param {Object} instance
+   */
+  verifyPlugins(instance) {
```
adamweeks: Let's update this to be `sparkInstance` too so we have consistency throughout the file.

packages/node_modules/@ciscospark/react-redux-spark/src/component.js
akoushke: nvm. the reason does not apply to my case.

packages/node_modules/@ciscospark/react-redux-spark/src/component.test.js
```diff
@@ -4,17 +4,82 @@ import {Provider} from 'react-redux';
...
       </Provider>
     );
 
-    expect(component).toMatchSnapshot();
+    expect(component.toTree()).toMatchSnapshot();
+  });
+
+  afterEach(() => {
+    sparkInstance = null;
+    sparkInstance = null;
```
adamweeks: Does this need to be done twice?

PR #980, opened 2019-06-17
chore(package): update description

adamweeks:


akoushke:
hahahaha. that change

PR #983, opened 2019-07-08
Upgrade sdk for s3 upload support

adamweeks:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-74784


adamweeks:
> Do you need an [upgrade doc like this](https://github.com/webex/webex-js-sdk/blob/master/UPGRADING.md) for developers?

@mattnorris good point, I'll add one in.

PR #998, opened 2019-09-12
Both clients need to leave the meeting before checking for checking for the Message Widget

taymoork2:


adamweeks:
img

adamweeks:
Rerunning circle tests since these still failed 😢

PR #999, opened 2019-09-19
Fixing tap tests for stability and hydra support

adamweeks:
Tap tests were failing due to them not being updated with the hydraid support for space destination. This PR converts the TAP tests to using the test user helper methods for space and user creation.

There are also some optimizations added due to animations not being disabled on the production server. This was causing click events to not register properly because the activity menu was not fully animated away before the click was attempted.

### DIFF
--- a/test/journeys/lib/test-helpers/space-widget/main.js
+++ b/test/journeys/lib/test-helpers/space-widget/main.js
@@ -29,6 +29,9 @@ export function switchToMessage(aBrowser) {
   }
   aBrowser.waitForVisible(elements.messageActivityButton);
   aBrowser.click(elements.messageActivityButton);
+
+  // Activity menu animates the hide, wait for it to be gone
+  aBrowser.waitForVisible(elements.activityMenu, 1500, true);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


test/journeys/specs/tap/widget-space/space.js
```diff
@@ -30,76 +29,14 @@ describe('Widget Space: Group Space: TAP', () => {
...
-        config: {
-          logger: {
-            level: 'error'
-          }
-        }
-      });
-
-      return lorraine.spark.internal.mercury.connect();
-    }));
+  before('create test users and spaces', () => {
```
InteractiveTimmy: Good job cleaning up these before/after cases.

test/journeys/lib/test-helpers/space-widget/main.js
```diff
@@ -29,6 +29,9 @@ export function switchToMessage(aBrowser) {
   }
   aBrowser.waitForVisible(elements.messageActivityButton);
   aBrowser.click(elements.messageActivityButton);
+
+  // Activity menu animates the hide, wait for it to be gone
```
InteractiveTimmy: These lines are rather helpful to determine test flow. Looks good!

PR #1000, opened 2019-09-30
Update vulnerable packages

adamweeks:
Completely removed the stylelint packages and configs since they don't work on files within `node_modules` (aka all of our monorepo files) and has been throwing security issues.

If we decide we want to lint our css, we will add that back in another way.

Also updated other packages


arun3528:
@adamweeks   Can you update the SDK to latest if possible .

adamweeks:
> @adamweeks Can you update the SDK to latest if possible .

@arun3528 Tim will be opening another PR to update to the latest SDK, but it needs some federation work which doesn't fit in this PR

PR #1007, opened 2019-10-08
fix(circle): version number for cdn

adamweeks:
Version grepping solution via https://gist.github.com/DarrenN/8c6a5b969481725a4413

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,7 +231,7 @@ jobs:
       - run:
           name: 'Grab Version Number'
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -231,7 +231,7 @@ jobs:
       - run:
           name: 'Grab Version Number'
           command: |
-            echo 'VERSION_NUMBER="$(npx -c \'echo "$npm_package_version"\')"' >> $BASH_ENV
+            echo VERSION_NUMBER=$(npx -c 'echo \"$npm_package_version\"') >> $BASH_ENV
```
InteractiveTimmy: Goot catch, works locally.

PR #1008, opened 2019-10-08
chore(circle): consistency updates

adamweeks:


.circleci/config.yml
```diff
@@ -70,17 +70,17 @@ jobs:
   build_for_tests:
     executor: main-executor
     environment:
-      - FEDERATION: "true"
```
adamweeks: Standardizing on the format listed here: https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-job

.circleci/config.yml
```diff
@@ -213,6 +213,7 @@ jobs:
       - persist_to_workspace:
           root: ~/react-widgets
           paths:
+            - package.json
```
adamweeks: Persisting the package.json file that has been updated with the latest version. Grabbing the version in the `deploy_to_cdn` was reporting the previous version.

.circleci/config.yml
```diff
@@ -70,17 +70,17 @@ jobs:
...
+      NODE_ENV: test
+      ACL_SERVICE_URL: https://acl-intb.ciscospark.com/acl/api/v1
+      ATLAS_SERVICE_URL: https://atlas-intb.ciscospark.com/admin/api/v1
+      ENCRYPTION_SERVICE_URL: https://encryption-intb.ciscospark.com/encryption/api/v1
+      IDENTITY_BASE_URL: https://identitybts.webex.com
+      CONVERSATION_SERVICE: https://conversation-intb.ciscospark.com/conversation/api/v1
+      HYDRA_SERVICE_URL: https://apialpha.ciscospark.com/v1/
+      IDBROKER_BASE_URL: https://idbrokerbts.webex.com
+      U2C_SERVICE_URL: https://u2c-intb.ciscospark.com/u2c/api/v1
+      WDM_SERVICE_URL: https://wdm-intb.ciscospark.com/wdm/api/v1
```
InteractiveTimmy: I like this code style change, much more consistent with the rest of the content.

.circleci/config.yml
```diff
@@ -223,15 +224,15 @@ jobs:
     executor: main-executor
     working_directory: ~/react-widgets
     environment:
-      - AWS_BUCKET: "code.s4d.io"
-      - AWS_REGION: "us-east-1"
+      AWS_BUCKET: code.s4d.io
```
taymoork2: I already put the bucket in the env on the dashboard
The region can also go there

adamweeks: @taymoork2 the bucket and region were not in the web config until this morning when I added it. I prefer them to be in the config file instead since they are not sensitive data.

PR #1013, opened 2019-10-10
test(journeys): add delay for test user propagation to smoke tests

adamweeks:
Adding more stability features to smoke tests that get run on PRs. 

Calling tests on Firefox run from circle to sauce are flaky at best. We have chosen to skip the calling tests for now until the meetings plugin can be utilized which gives better firefox support.


adamweeks:
> LGTM
> 
> I would move the pauses into a sleep in the `setupGroupTestUsers()`

Looks like we already did that actually! I've removed the extra sleeps.

PR #1017, opened 2019-10-22
Rename CDN links from latest+prerelease to alpha and beta

taymoork2:
User alpha/beta for latest and prerelease


.circleci/config.yml
```diff
@@ -313,9 +347,12 @@ workflows:
           filters:
               branches:
                 only: master
-      - version_and_publish:
+      - rename_latest_to_alpha:
```
adamweeks: I'm assuming we'll remove this in a following PR?

taymoork2: Yup

PR #1018, opened 2019-10-22
chore(circleci): remove rename job

taymoork2:


.circleci/config.yml
```diff
@@ -347,12 +330,9 @@ workflows:
           filters:
               branches:
                 only: master
-      - rename_latest_to_alpha:
-          requires:
-              - install
       - version_and_publish:
           requires:
-              - rename_latest_to_alpha
+            - install
```
adamweeks: White space nit pick, can you align the start of this with the other requires around it

taymoork2: I ran whatever formatter my vscode has setup for yaml (I'm guessing prettier) and this is what I got

PR #1022, opened 2019-10-30
Version file

taymoork2:


.circleci/config.yml
```diff
@@ -293,6 +302,30 @@ jobs:
...
+    executor: main-executor
+    working_directory: ~/react-widgets
+    steps:
+      - checkout
+      - run: git remote add upstream git@github.com:webex/react-widgets.git
+      - run: git fetch --all --tags --prune
+      - run:
+          name: 'Grab Version Number'
+          command: |
+            echo export VERSION_NUMBER=$(curl 'https://code.s4d.io/widget-demo/production/.version') >> $BASH_ENV
```
adamweeks: If the curl fails here, does this command fail gracefully?

taymoork2: nope
You got a quick fix?

.circleci/config.yml
```diff
@@ -173,21 +173,21 @@ jobs:
       - run:
           name: 'Generate Version Number'
           command: |
-            npm run release -- --release-as patch --no-verify
-            echo export VERSION=$(npx -c 'echo \"$npm_package_version\"') >> $BASH_ENV
+            npm run release -- --release-as patch --prerelease alpha --no-verify
```
adamweeks: please revert this line, we don't want to release as alpha version.

.circleci/config.yml
adamweeks: The only thing I can think of is just separating the curl command out and making it error if it fails.

adamweeks: Could you rename the variable to `PRODUCTION_VERSION_NUMBER` so that it is different from the `VERSION_NUMBER` used in the previous jobs.

.circleci/config.yml
```diff
@@ -293,6 +302,30 @@ jobs:
...
+          command: |
+            echo export VERSION_NUMBER=$(curl 'https://code.s4d.io/widget-demo/production/.version') >> $BASH_ENV
+      - run:
+          name: 'Delete old production tag'
+          command: git push upstream :refs/tags/production
+      - run: git checkout tags/v${VERSION_NUMBER}
+      - run:
+          name: 'Create new production tag'
+          command: |
+            git tag -f production
```
adamweeks: img

.circleci/config.yml
```diff
@@ -293,6 +302,30 @@ jobs:
...
+    executor: main-executor
+    working_directory: ~/react-widgets
+    steps:
+      - checkout
+      - run: git remote add upstream git@github.com:webex/react-widgets.git
+      - run: git fetch --all --tags --prune
+      - run:
+          name: 'Grab Version Number'
+          command: |
+            echo export VERSION_NUMBER=$(curl -f 'https://code.s4d.io/widget-demo/production/.version') >> $BASH_ENV
```
adamweeks: If curl fails with a 403 here, it is just going to assign VERSION_NUMBER to 403, not really fail.

PR #1023, opened 2019-10-30
chore(sri): add support for env var private keys

adamweeks:
The upgrade to circle CI pipeline had the drawback of not having the ability to store secret files like the private key. 

This change allows us to store the contents of a private key in the PRIVATE_KEY env var, which has now been added on the circle ci config.


adamweeks:
> Does the circle config not need updating?

@taymoork2 the config just needed the PRIVATE_KEY ENV var added, which it has. 
The script was already attempting to build the sri but failing (silently): https://github.com/webex/react-widgets/blob/master/.circleci/config.yml#L185

PR #1025, opened 2019-10-31
fix(adaptive-card): Message contains files but not cards

mokandre:


packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
```diff
@@ -110,7 +110,7 @@ function ActivityItem(props) {
       itemComponent = (
         <ActivityShareFiles
           renderAdaptiveCard={(shouldRenderAdaptiveCard && adaptiveCardFeatureState === FEATURES_WIDGET_ADAPTIVE_CARD_ON
-            ) ? shouldRenderAdaptiveCard : !shouldRenderAdaptiveCard}
+          )}
```
InteractiveTimmy: Good catch!

PR #1029, opened 2019-11-14
Move tap test to CircleCI

taymoork2:


.circleci/config.yml
```diff
@@ -319,6 +319,29 @@ jobs:
           name: 'Push tag to upstream/master'
           command: git push --tags upstream
 
+  tap_test:
+    executor: main-executor
+    environment:
+      SAUCE: true
+    steps:
+      - checkout
```
adamweeks: We will need to checkout the `production` tag. The rest seems ok, were you able to run this on circle to test?

PR #1030, opened 2019-11-18
feat(ac-markdown)Support markdown for adaptive cards

mokandre:
Description of changes made (Required)
* Added  function for handing markdown by using an existing function
* Added unit test 

https://www.npmjs.com/package/adaptivecards#supporting-markdown 

Associated issue number/tracking link (Required)
[94103](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-94103)

**The following criteria should be met (Required)**
- [X] [commit guidelines](https://sqbu-github.cisco.com/WebExSquared/web-client/blob/master/CONTRIBUTING.md#github-commit-guidelines) were followed
- [X] Tests have been added for new functionality and all existing tests pass.
- [X] [pushing guidelines](https://sqbu-github.cisco.com/WebExSquared/web-client/blob/master/CONTRIBUTING.md#pushing-to-the-pipeline) are followed

Reviewers (Required)
@InteractiveTimmy @akoushke @adamweeks

### DIFF
--- a/packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
+++ b/packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
@@ -2,8 +2,30 @@ import * as AdaptiveCard from 'adaptivecards';
 import MDTextInput from '@ciscospark/react-component-md-text-input';
 import MDToggleInput from '@ciscospark/react-component-md-toggle-input';
 import MDChoiceSetInput from '@ciscospark/react-component-md-choiceset-input';
+import {parseMarkdown} from '@ciscospark/redux-module-activity';
 
 import {CARD_ATTACHMENT_TYPE, TEXT_INPUT_ELEMENT, TOGGLE_INPUT_ELEMENT, CHOICE_SET_INPUT_ELEMENT} from './constants';
+
+/* eslint no-param-reassign: 2 */
+// eslint-disable-next-line valid-jsdoc
+/**
+ * It used to handle markdown
+ * @param {string} text
+ * @returns {object}
+ */
+AdaptiveCard.AdaptiveCard.onProcessMarkdown = function processMarkdown(text, result = {}) {
+  try {
+    const markedString = parseMarkdown(text);
+
+    result.outputHtml = markedString !== null ? markedString : text;
+    result.didProcess = true;
+  }
+  catch (err) {
+    result.didProcess = false;
+  }
+
+  return result;
+};
 
 /**
  * It used to call momentum ui compoennts and returns patched adaptive card
--- a/packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
+++ b/packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
@@ -6,25 +6,25 @@ import {createMessageObject} from '@ciscospark/redux-module-activity';
 
 import {CARD_ATTACHMENT_TYPE, TEXT_INPUT_ELEMENT, TOGGLE_INPUT_ELEMENT, CHOICE_SET_INPUT_ELEMENT} from './constants';
 
+/* eslint no-param-reassign: 2 */
 /**
  * It used to handle markdown
+ * esLint is disabled for param-reassign because based on this documentation https://github.com/microsoft/AdaptiveCards/blob/master/source/nodejs/adaptivecards/README.md#supporting-markdown
+ * need to modify the result object only
  * @param {string} text
  * @param {object} result
  * @returns {object}
  */
 AdaptiveCard.AdaptiveCard.onProcessMarkdown = function processMarkdown(text, result = {}) {
-  const computedResult = {...result};
-
   try {
     const markedString = createMessageObject(text).content;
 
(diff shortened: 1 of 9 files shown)


packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
```diff
@@ -2,9 +2,31 @@ import * as AdaptiveCard from 'adaptivecards';
 import MDTextInput from '@ciscospark/react-component-md-text-input';
 import MDToggleInput from '@ciscospark/react-component-md-toggle-input';
 import MDChoiceSetInput from '@ciscospark/react-component-md-choiceset-input';
+import {parseMarkdown} from '@ciscospark/redux-module-activity';
 
 import {CARD_ATTACHMENT_TYPE, TEXT_INPUT_ELEMENT, TOGGLE_INPUT_ELEMENT, CHOICE_SET_INPUT_ELEMENT} from './constants';
 
+/* eslint no-param-reassign: 2 */
```
mokandre: esLinit is disabled at a functional level because it's not allowing to assign values to params

adamweeks: There's no need to disable this rule. You can create a new object instead of modifying the param object.

packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
```diff
@@ -2,9 +2,31 @@ import * as AdaptiveCard from 'adaptivecards';
...
+// eslint-disable-next-line valid-jsdoc
```
adamweeks: If you add the docs for the `result` param, you won't need to disable the eslint.

packages/node_modules/@ciscospark/redux-module-activity/src/actions.js
```diff
@@ -285,3 +285,9 @@ export function retryFailedActivity(failedActivity, spark) {
       .catch((error) => dispatch(submitActivityFailure(failedActivity, error)));
   };
 }
+
+export function parseMarkdown(text) {
+  const activityObject = createMessageObject(text);
```
adamweeks: Instead of adding this in the actions, lets just export the `createMessageObject` function in index.js:

`packages/node_modules/@ciscospark/redux-module-activity/src/index.js`:
```js
export {createMessageObject} from './helpers';
```

packages/node_modules/@ciscospark/redux-module-activity/src/actions.test.js
```diff
@@ -229,3 +229,11 @@ describe('redux-module-activity actions', () => {
     });
   });
 });
+
+describe('parseMarkdown', () => {
```
adamweeks: If we export `createMessageObject` we can remove this test.

adamweeks:
@mokandre could you please rebase this on the latest master as well, we have some pipeline changes that need to take effect.

packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
mokandre: created a new object and assigned a new object to the result again. Removed  esLint rule

packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
mokandre: done

packages/node_modules/@ciscospark/react-component-utils/src/adaptiveCardUtil.js
```diff
@@ -2,9 +2,33 @@ import * as AdaptiveCard from 'adaptivecards';
...
+  try {
+    const markedString = createMessageObject(text).content;
+
+    computedResult.outputHtml = markedString !== null ? markedString : text;
+    computedResult.didProcess = true;
+  }
+  catch (err) {
+    computedResult.didProcess = false;
+  }
+  Object.assign(result, computedResult);
```
akoushke: any reason we are not returning the `computedResult` instead of assigning it to the parameter and return the parameter?

akoushke: you are copying from the results and copying back to results. its looks a bit redundant to me.

mokandre: @akoushke, markdown is not working when I'm returning `computedResult` so that's reason again I assigned it to result

adamweeks: It looks like we have to actually modify the `result` object based on the documentation: https://github.com/microsoft/AdaptiveCards/blob/master/source/nodejs/adaptivecards/README.md#supporting-markdown

In this case, let's go back to disabling the property change eslint rule, but just for this function. We just ask that you include the link to the docs and an explanation of why the eslint rule was disabled.

mokandre: disabled the property

PR #1014, opened 2019-10-16
feat : Render adaptive card behind feature toggle

mokandre:
Description of changes made 
* Feature toggle for adaptive cards
* Rendering adaptive card
* Unit test case coverage

Associated issue number/tracking link 

[SPARK-94093](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-94093)
[SPARK-97152](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-97152)

Reviewers 
@adamweeks

PR #1034, opened 2019-12-04
feat(ac-images):rehosting images

mokandre:
**Description of changes made (Required)**
- Changes in adaptive card component
- Added a function to replace index with URL's

**Associated issue number/tracking link (Required)**
[94105](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-94105)

**The following criteria should be met (Required)**
- [X] [commit guidelines](https://sqbu-github.cisco.com/WebExSquared/web-client/blob/master/CONTRIBUTING.md#github-commit-guidelines) were followed
- [X] Tests have been added for new functionality and all existing tests pass.
- [X] [pushing guidelines](https://sqbu-github.cisco.com/WebExSquared/web-client/blob/master/CONTRIBUTING.md#pushing-to-the-pipeline) are followed

**Reviewers (Required)**
  @adamweeks


mokandre:
@lalli-flores journey tests for image rehosting coming under different task itself:https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-108063 .Still, Do I need to raise as draft PR?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
...
-        getAdaptiveCard(this.props.cards, this.props.displayName, this.props.sparkInstance, this.addChildNode)
+        getAdaptiveCard(this.state.cards, this.props.displayName, this.props.sparkInstance, this.addChildNode)
       );
     }
   }
 
+  static getDerivedStateFromProps(nextProps) {
+    if (nextProps.verb === API_ACTIVITY_VERB.SHARE) {
+      try {
+        const decryptedURLs = nextProps.items
```
lalli-flores: Just trying to understand the code here, where does `verb` and `items` come from?
My understanding is that `nextProps` represent the component's props, but from where i can see those are only `cards`, `displayname` and `sparkInstance`

(side note - we shouldn't propagate the name "spark" and we should use "webex")

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
...
+          .filter((file) => file.type === CARD_CONTAINS_IMAGE)
+          .map((file) => {
+            const thumbnail = file.mimeType === 'image/gif' ? nextProps.share.getIn(['files', file.url]) : nextProps.share.getIn(['files', file.image.url]);
+
+            if (thumbnail) {
```
lalli-flores: Where would it be a case where there is no `thumbnail`?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
...
     }
   }
 
+  static getDerivedStateFromProps(nextProps) {
+    if (nextProps.verb === API_ACTIVITY_VERB.SHARE) {
+      try {
+        const decryptedURLs = nextProps.items
+          .filter((file) => file.type === CARD_CONTAINS_IMAGE)
+          .map((file) => {
+            const thumbnail = file.mimeType === 'image/gif' ? nextProps.share.getIn(['files', file.url]) : nextProps.share.getIn(['files', file.image.url]);
```
lalli-flores: What does `getIn` return?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
...
+            if (thumbnail) {
+              const objectUrl = thumbnail.get('objectUrl');
+
+              return objectUrl;
+            }
+
+            return undefined;
+          });
+        const cards = JSON.parse(nextProps.cards[0]);
+        const undefinedUrls = decryptedURLs.filter((file) => file === undefined);
```
lalli-flores: What's the purpose of returning `undefined` earlier if they're going to be filtered out? wouldn't it be better to just not add them (by not returning a few lines earlier) to begin with?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
...
+
+        if (undefinedUrls.length === 0) {
+          const urlTypes = ['url', 'backgroundImage', 'iconUrl'];
```
lalli-flores: Is there a reason why this are passed in, instead just handled within the utility function?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
...
+          const replacedImagesinCards = replaceIndexWithBlobURL(urlTypes, cards, decryptedURLs);
```
lalli-flores: Is there a reason why the `decryptedURLs` aren't calculated within `replaceIndexWithBlobURL`? I don't see anything on the calculation of the URL that's dependent on the component code.
It's a bit hard to follow the logic when half of it it's done in the component (the URL calculation) and the other half it's done in a helper function (the attribute replacement)

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -1,14 +1,18 @@
...
+import {getAdaptiveCard, CARD_CONTAINS_IMAGE, replaceIndexWithBlobURL, API_ACTIVITY_VERB} from '@ciscospark/react-component-utils';
 
 class AdaptiveCard extends React.Component {
   constructor(props) {
     super(props);
     this.nodeElement = React.createRef();
     this.state = {
-      childNodes: []
+      childNodes: [],
+      hasReplacedImagesInJSON: false,
```
lalli-flores: I don't think this should be a component state. There is only one place that this is being set (line 54) and never updated again

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
   componentDidMount() {
     if (Object.prototype.hasOwnProperty.call(this, 'nodeElement') && Object.prototype.hasOwnProperty.call(this.nodeElement, 'current')) {
       this.nodeElement.current.appendChild(
-        getAdaptiveCard(this.props.cards, this.props.displayName, this.props.sparkInstance, this.addChildNode)
+        getAdaptiveCard(this.state.cards, this.props.displayName, this.props.sparkInstance, this.addChildNode)
       );
     }
   }
 
+  static getDerivedStateFromProps(nextProps) {
```
lalli-flores: Can you explain the reason for using this lifecyle? [There's a lot of warnings from React team](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html) on using this lifecycle.

From my understanding of the code, it seem like this is only expected to run once (because of the `hasReplacedImagesInJSON` set once) but correct me if I'm misinterpreting

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -1,14 +1,18 @@
...
 class AdaptiveCard extends React.Component {
   constructor(props) {
     super(props);
     this.nodeElement = React.createRef();
     this.state = {
-      childNodes: []
+      childNodes: [],
+      hasReplacedImagesInJSON: false,
+      hasReplacedCard: false,
+      cards: this.props.cards
```
lalli-flores: Why are you setting a prop as a state?

packages/node_modules/@ciscospark/react-component-activity-share-files/src/index.js
```diff
@@ -101,7 +106,14 @@ export function ActivityShareFiles(props) {
   return (
     <ActivityItemBase {...props}>
       <div className={classNames('ciscospark-activity-share-list', styles.shareList)}>
-        {renderAdaptiveCard && <AdaptiveCard cards={cards} displayName={displayName} />}
+        {renderAdaptiveCard && <AdaptiveCard
+          cards={cards}
+          displayName={displayName}
+          items={files}
+          verb={verb}
```
lalli-flores: Is this component the one [defined in this file](https://github.com/webex/react-widgets/blob/5b1a635ef2b1e5eb25e53cf1b45ef0330dcb3df6/packages/node_modules/%40ciscospark/react-component-adaptive-card/src/index.js)? Because looking at the [props defined](https://github.com/webex/react-widgets/blob/5b1a635ef2b1e5eb25e53cf1b45ef0330dcb3df6/packages/node_modules/%40ciscospark/react-component-adaptive-card/src/index.js#L102) they don't seem to be matching with what is being called here

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable radix */
+/* eslint-disable import/prefer-default-export */
+/**
+ * replaces the index with the uri in cards
+ * @param {Array} urlTypes
+ * @param {Object} obj
+ * @param {Array} items
+ * @returns {Object}
+ */
+export function replaceIndexWithBlobURL(urlTypes, obj, items) {
```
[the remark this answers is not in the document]
mokandre: esLint is throwing an error(no-param-reassign) if trying to assign to params so tried to copy the object and modifying the copied object but changed the obj to descriptive name

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
```diff
@@ -0,0 +1,30 @@
...
+  const activityObject = obj;
+
+  if (activityObject !== null && activityObject !== undefined) {
+    Object.keys(activityObject).forEach((property) => {
+      if (Object.prototype.hasOwnProperty.call(activityObject, property)) {
+        const index = urlTypes.indexOf(property);
+        const key = urlTypes[index];
+
+        if (property === key && !Number.isNaN(parseInt(activityObject[key]))) {
+          activityObject[key] = items[parseInt(activityObject[key])];
```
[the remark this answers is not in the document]
mokandre: Modifying the copied object (refer line 10)

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
mokandre: items in the sense it contains decryptedURLs which need to replace with the index. Changed the name of items to decryptedURLs

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable radix */
+/* eslint-disable import/prefer-default-export */
```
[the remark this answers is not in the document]
mokandre: removed disabling esLint but added code for not throwing esLint errors

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
mokandre: Getting the key and index of URL types and using it to assign for activityObject (@ line 18)

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: It gives the blob URL which required to replace in the place of the index

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: This condition is for avoiding the error if trying to get objectURL when there is no content in `thumbnail`

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Initially, there is a chance to get the URL Value as undefined instead of the index. So filtering the undefinedUrls and  if there is no content in thumbnail also arrow function will expect a return value

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: State variable setting happens every time when props updated. This state variable is using in the condition of render to replace the child node(at line 83).

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: As I mentioned earlier here:https://github.com/webex/react-widgets/pull/1034#discussion_r357017415 .   `getDerivedStateFromProps` will execute when props updated.

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,53 @@ class AdaptiveCard extends React.Component {
...
+        const cards = JSON.parse(nextProps.cards[0]);
+        const undefinedUrls = decryptedURLs.filter((file) => file === undefined);
+
+        if (undefinedUrls.length === 0) {
+          const urlTypes = ['url', 'backgroundImage', 'iconUrl'];
+          const replacedImagesinCards = replaceIndexWithBlobURL(urlTypes, cards, decryptedURLs);
+
+          cards[0] = JSON.stringify(replacedImagesinCards);
+
+          return {hasReplacedImagesInJSON: true, cards};
```
mokandre: This will set the state to cards and I'm passing the updated cards to getAdaptiveCard function line 86

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Please Refer:https://github.com/webex/react-widgets/pull/1034/files/5b1a635ef2b1e5eb25e53cf1b45ef0330dcb3df6#r357024069

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: There is no particular reason,`replaceIndexWithBlobURL` function is only for replacing the value so remaining code for URL calculation is in the component

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: moved into helper function

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: [Adaptive card component](https://github.com/webex/react-widgets/blob/5b1a635ef2b1e5eb25e53cf1b45ef0330dcb3df6/packages/node_modules/%40ciscospark/react-component-adaptive-card/src/index.js) calling from two places.One [component](https://github.com/webex/react-widgets/blob/5b1a635ef2b1e5eb25e53cf1b45ef0330dcb3df6/packages/node_modules/%40ciscospark/react-component-activity-item/src/index.js)  is passing three props and
 [share component](https://github.com/webex/react-widgets/blob/5b1a635ef2b1e5eb25e53cf1b45ef0330dcb3df6/packages/node_modules/%40ciscospark/react-component-activity-share-files/src/index.js) need to pass more props. In my use case, I'm getting the required props from a [share component](https://github.com/webex/react-widgets/blob/5b1a635ef2b1e5eb25e53cf1b45ef0330dcb3df6/packages/node_modules/%40ciscospark/react-component-activity-share-files/src/index.js) as `nextProps`. Though those are defined  in PropTypes it will throw an error because `this.props` is not using anywhere for remaining `verb`, `items` and `share`.

packages/node_modules/@ciscospark/react-component-activity-share-files/src/index.js
mokandre: Refer:https://github.com/webex/react-widgets/pull/1034#discussion_r356935122

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
mokandre: added test case

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
lalli-flores: I think the URL calculation logic should be moved to `replaceIndexWithBlobURL`, after all the function name already describes that it will change index => URL.

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
lalli-flores:
```js
const activityObject = cardsActivity;
```
does _**not**_ copy an object, it just creates a new reference variable, and hence why you're getting the error

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: `replaceIndexWithBlobURL` is a recursive function.If we move URL calculation to  `replaceIndexWithBlobURL` then unnecessarily code will execute multiple times

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
mokandre: Screenshot 2019-12-13 at 8 21 00 AM
That's the reason for creating new object and modifying it

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
lalli-flores: You can store the cards as an instance variables of the class. The states should only be things that render should be "listening to" to update the display. This does not seem the case for `cards`

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
lalli-flores: Did you read the [alternatives to `getDeriveStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops)? Would the suggestion of using `componentDidUpdate` be sufficient?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
lalli-flores: Is this something that can be done "just in time" within `replaceIndexWithBlobURL`?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -17,11 +21,52 @@ class AdaptiveCard extends React.Component {
...
+
+            if (thumbnail) {
+              const objectUrl = thumbnail.get('objectUrl');
+
+              return objectUrl;
+            }
+
+            return undefined;
+          });
+        const cards = JSON.parse(nextProps.cards[0]);
```
lalli-flores: Why are you looking only at the first card?

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
```diff
@@ -0,0 +1,31 @@
+const urlTypes = ['url', 'backgroundImage', 'iconUrl'];
+
+/**
+ * replaces the index with the uri in cards
+ * @param {Object} cardsActivity
+ * @param {Array} decryptedURLs
+ * @returns {Object}
+ */
+export function replaceIndexWithBlobURL(cardsActivity, decryptedURLs) {
```
lalli-flores: If `cardsActivity` is an object with properties and `decryptedURLs` is an array of strings, how do you know what URL corresponds to what object/property?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Cards are being consumed by render method to rerender after updating the images into the JSON.

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
```diff
@@ -0,0 +1,31 @@
...
+  const activityObject = cardsActivity;
+
+  if (activityObject) {
+    Object.keys(activityObject).forEach((property) => {
```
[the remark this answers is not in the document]
mokandre: Screenshot 2019-12-13 at 11 32 35 AM
esLint is not allowing to use `for in` .Suggesting to use Object.keys itslef

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Screenshot 2019-12-13 at 12 14 52 PM
Though default values for props are added, esLint is throwing an error . Can I disable the esLint here?

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
mokandre: storing the blob URL at the same index as what the URL value of the card object is pointing to

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: decided to render only one card for now

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
lalli-flores: The linter is recommending to use it [like this](https://stackoverflow.com/a/4166579)

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
lalli-flores: Im saying that in you only do the calculation once you get to the base condition, instead of pre-calculating everything, running multiple times through the card (you're traversing the cards once per filter operation)

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
mokandre: My bad, I put it in the wrong way https://github.com/webex/react-widgets/pull/1034#discussion_r356979786). A reference variable is enough to modify so I'm not copying into other object

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: added propTypes and defaultProps

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: used `componentDidUpdate` lifecycle

packages/node_modules/@ciscospark/react-component-utils/src/ReplaceIndexWithBlobUrl.js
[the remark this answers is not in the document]
mokandre: replaced with for in but added esLint throwing error as `next-line no-restricted-syntax` so disabled at that line

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Tried to use the `componentDidUpdate' life cycle method but props of last activity is not getting updated

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Screenshot 2019-12-13 at 12 14 52 PM
So can I add propTypes and disable esLint?

mokandre: added `componentDidUpdate` and added extra props in propTypes.I have to set state for re-render.

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: There is no base condition for URL calculation as we required updated `props` every time

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -22,6 +26,36 @@ class AdaptiveCard extends React.Component {
...
+    this.state.childNodes.forEach((childNode) => {
+      ReactDOM.unmountComponentAtNode(childNode);
+    });
+  }
+
+  /**
+   * set state to tohasReplacedImagesInJSON
+   * @returns {void}
+   */
+  setupState() {
```
lalli-flores: Is this called more than in one place? does it need to be a method?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -46,13 +90,15 @@ class AdaptiveCard extends React.Component {
 AdaptiveCard.propTypes = {
   cards: PropTypes.array,
   displayName: PropTypes.string,
-  sparkInstance: PropTypes.object
+  sparkInstance: PropTypes.object,
+  verb: PropTypes.string
 };
 
 AdaptiveCard.defaultProps = {
   cards: [],
```
lalli-flores: This component is supposed to represent one card, so why are you making this prop an array, if you also only look at the first element only?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -1,17 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import ReactDOM from 'react-dom';
 import classnames from 'classnames';
-import {getAdaptiveCard} from '@ciscospark/react-component-utils';
+import {getAdaptiveCard, getDecryptedURls, API_ACTIVITY_VERB} from '@ciscospark/react-component-utils';
```
lalli-flores: Can you make this a relative import? we should avoid potential circular dependencies

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
```diff
@@ -22,6 +26,36 @@ class AdaptiveCard extends React.Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps !== this.props && this.props.verb === API_ACTIVITY_VERB.SHARE) {
```
lalli-flores: Can you add a snapshot test for an adaptive card with blob URLs to tests this changes?

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: we can not set the state in `componentDidUpdate` lifecycle component so that's the reason for setting the state in method

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: If we need to to render multiple cards then no need to change parent component props as it is representing array of cards already

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Please refer this https://github.com/webex/react-widgets/pull/1014#discussion_r339013901. As discussed with @adamweeks if we are not able to do in snapshot testing , will check in journey tests

packages/node_modules/@ciscospark/react-component-utils/src/decryptImageUrl.test.js
```diff
@@ -0,0 +1,47 @@
+import {replaceIndexWithBlobURL} from './decryptImageUrl';
+
+describe('helper functions', () => {
```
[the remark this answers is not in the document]
mokandre: Refer:https://github.com/webex/react-widgets/pull/1034#discussion_r359208467 .As we discussed it's need props from redux store

packages/node_modules/@ciscospark/react-component-utils/src/decryptImageUrl.js
```diff
@@ -0,0 +1,62 @@
...
+const urlTypes = ['url', 'backgroundImage', 'iconUrl'];
+
+/**
+ * replaces the index with the uri in cards
+ * @param {Object} cardsActivity
+ * @param {Array} decryptedURLs
+ * @returns {Object}
+ */
+export function replaceIndexWithBlobURL(cardsActivity, decryptedURLs) {
+  const activityObject = cardsActivity;
```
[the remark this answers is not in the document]
mokandre: @lalli-flores I tried with cloneDeep as well but it's not working

packages/node_modules/@ciscospark/react-component-adaptive-card/src/index.js
mokandre: Screenshot 2019-12-18 at 4 14 20 PM
While I'm running test cases. It's throwing an error as it's expecting in a particular format.

packages/node_modules/@ciscospark/react-component-utils/src/decryptImageUrl.test.js
```diff
@@ -0,0 +1,47 @@
...
+          items: [
+            {
+              type: 'Image',
+              url: '0'
+            }
+          ]
+        }
+
+        ]
+      }]
```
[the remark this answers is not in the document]
InteractiveTimmy: Thank you for addressing this.

PR #1037, opened 2019-12-12
FIX: Widget Space Email Setup

adamweeks:
When using a destination type of email and the destination ID is mixed case, convert to lowercase to match the response from the server.

Also added a missing test suite around the `setup.js` file of the space widget. This caused me to have to move the constants into its own file due to jest trying to load reselect when doing `import X from './'`.

Commits should be separated and easy to follow.

### DIFF
--- a/packages/node_modules/@ciscospark/widget-space/src/enhancers/setup.js
+++ b/packages/node_modules/@ciscospark/widget-space/src/enhancers/setup.js
@@ -96,10 +96,15 @@ function setup(props, prevProps) {
           destinationId,
           destinationType
         } = props;
+        let id = destinationId;
+
+        if (destinationType === destinationTypes.EMAIL && destinationId) {
+          id = destinationId.toLowerCase();
+        }
 
         props.storeDestination({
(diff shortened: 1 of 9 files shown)


packages/node_modules/@ciscospark/widget-space/src/enhancers/setup.js
```diff
@@ -64,7 +64,7 @@ function setup(props, prevProps) {
     if (!mercuryStatus.hasConnected
       && !mercuryStatus.connecting
```
taymoork2: I don't know why, but I hate this syntax

adamweeks: Open to suggestions for change. One thing I thought about was to make it a function or variable so it is easier to read like:

```js
const shouldConnectToMercury = !mercuryStatus.hasConnected && !mercuryStatus.connecting && !mercuryStatus.connected && sparkState.registered;

if (shouldConnectToMercury) {
```

PR #1038, opened 2019-12-18
Use node lts/dubnium (v10)

taymoork2:


package.json
```diff
@@ -93,6 +93,7 @@
     "ampersand-events": "^2.0.2",
     "babel-polyfill": "^6.26.0",
     "bowser": "^1.9.4",
+    "builtin-modules": "^3.1.0",
```
adamweeks: What is this for?

adamweeks:
Could you investigate the linting and test failures?

package.json
taymoork2: for some reason, [this](https://github.com/webex/react-widgets/blob/master/scripts/utils/deps.js#L6) started failing whenever I tried to build

taymoork2: this was an issue with wiping `node_modules` and `package-lock` and "starting fresh" with a `npm i`
Reverting back to the `package-lock` in the master branch and running `npm ci` corrected the issue(s)

adamweeks: Gotcha, if it is required, let's separate it into another commit with the message `fix(package): add builtin-modules package`.

taymoork2:
eslint issues (before the revert) were pop up during jest runs
The updated `package-lock` brought these issues forward when eslint was run
```
Warning: Failed prop type: The prop `actor` is marked as required in `FileShareDisplay`, but its value is `undefined`.
        in FileShareDisplay
```

PR #1040, opened 2020-01-06
fix(r-c-utils): update for federation

mwegman:
- update validateAndDecodeId to use SDK
- use 'url' to load previous messages

### DIFF
--- a/packages/node_modules/@ciscospark/react-component-utils/src/uuid-utils.js
+++ b/packages/node_modules/@ciscospark/react-component-utils/src/uuid-utils.js
@@ -1,4 +1,4 @@
-import {base64} from '@webex/common';
+import {base64, deconstructHydraId} from '@webex/common';
 
 /**
  * Tests if a string is a UUID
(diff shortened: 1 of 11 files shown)


packages/node_modules/@ciscospark/react-component-utils/src/uuid-utils.js
```diff
@@ -33,31 +33,31 @@ export function getConvoIdFromUrl(convoUrl) {
...
-      resultId = base64uuid;
+  }
+  catch (e) {
+    if (e.name !== 'InvalidCharacterError') {
+      throw e;
     }
   }
 
-  return resultId;
+  return {};
```
adamweeks: This is no longer returning a false when not verified. Does this affect anything using it downstream? 

Also, could you update the jsdoc to match what the new return value is?

packages/node_modules/@ciscospark/react-component-utils/src/uuid-utils.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+    it('deconstructs a hydra ID', () => {
+      const id = uuid.v4();
+      const cluster = 'us';
+      const hydraId = buildHydraRoomId(id, cluster);
+
+      expect(validateAndDecodeId(hydraId)).toEqual({id, cluster, type: hydraTypes.ROOM});
+    });
+
+    [null, '', 'AB%CD']
```
adamweeks: Could you separate these into three `it` blocks? We forgot to mention it during our scrum meetings, but I believe we didn't want the test looping (am I correct @lalli-flores?).

packages/node_modules/@ciscospark/react-component-utils/src/uuid-utils.test.js
```diff
@@ -0,0 +1,36 @@
+import uuid from 'uuid';
+import {buildHydraRoomId, hydraTypes} from '@webex/common';
+
+import {validateAndDecodeId} from './uuid-utils';
+
+describe('uuid-utils', () => {
```
adamweeks: Thanks for adding these!

packages/node_modules/@ciscospark/react-component-utils/src/uuid-utils.js
mwegman: The rest of this PR are the changes necessary to move from 'false' to an empty object.

PR #1043, opened 2020-01-14
Migrate packages to @webex

adamweeks:
This PR migrates all the packages to the @webex namespace. Another PR will follow deprecating the original @ciscospark packages.


adamweeks:
Note: netlify is failing to deploy because it is still using the old @ciscospark package, once this merges, we will update the config.

PR #1044, opened 2020-01-14
fix(federation): use 'url' to fetch a space

mwegman:


mwegman:
SPARK-114799 #done

packages/node_modules/@ciscospark/redux-module-spaces/src/actions.js
```diff
@@ -171,27 +171,25 @@ export function updateSpaceWithActivity(activity, isSelf, isReadable = false) {
  *
  * @export
  * @param {Object} sparkInstance
- * @param {String} spaceId
+ * @param {String} space
```
adamweeks: `space` is expected to be an object, correct?

packages/node_modules/@ciscospark/widget-space/src/actions.js
```diff
@@ -154,7 +154,7 @@ export function getSpaceDetails({
 
     return sparkInstance.request({
       service: 'hydra',
-      resource: `/rooms/${spaceIdUUID}`
+      resource: `/rooms/${destinationId}`
```
adamweeks: I thought you said this doesn't work with a request to `/rooms/<hydra encoded url>`?

mwegman: It doesn't work if it is a non 'US' cluster.

PR #1045, opened 2020-01-15
Migrate packages to @webex org

adamweeks:
This PR migrates all the packages to the @webex namespace. Another PR will follow deprecating the original @ciscospark packages.

Note: netlify is failing to deploy because it is still using the old @ciscospark package, once this merges, we will update the config.


scripts/utils/package.js
```diff
@@ -63,27 +63,15 @@ function getPackage(pkg) {
...
-  const sparkFullPath = path.resolve('packages/node_modules/@ciscospark', pkg);
   const webexFullPath = path.resolve('packages/node_modules/@webex', pkg);
   const fullPathWithPrefixExists = isDirectory(fullPathWithPrefix);
-  const sparkPathExists = isDirectory(sparkFullPath);
   const webexPathExists = isDirectory(webexFullPath);
 
   if (fullPathWithPrefixExists) {
     calculatedPackagesDir = fullPathWithPrefix;
   }
-  else {
```
adamweeks: We no longer want to auto-resolve the `@ciscospark` packages.

packages/node_modules/@webex/spark-widget-base/src/enhancers/withBrowserGlobals.js
```diff
@@ -0,0 +1,202 @@
...
+
+    return false;
+  }
+
+  return selectWidget(el);
+}
+
+
+function setupWidgetNamespace() {
+  if (!window.webex) {
```
adamweeks: Setting the global namespace to `webex` here instead of `ciscospark`.

packages/node_modules/@webex/spark-widget-base/src/enhancers/withBrowserGlobals.js
```diff
@@ -0,0 +1,202 @@
...
+    window.webex.widgetStore = {};
+  }
+  if (!window.webex.widgetFn) {
+    window.webex.widgetFn = {};
+  }
+  if (!window.webex.widget || typeof window.webex.widget !== 'function') {
+    window.webex.widget = getWidget;
+  }
+  // Legacy Support
+  if (!window.ciscospark) {
```
adamweeks: Adding the `ciscospark` global object for legacy support.

scripts/utils/package.js
InteractiveTimmy: Acknowledged, looks good.

PR #1046, opened 2020-01-15
'Webex' browser global namespace

adamweeks:
Builds off of #1045 and should be rebased once merged in. 

Changes the `window.ciscospark` implementation to `window.webex`.

### DIFF
--- a/packages/node_modules/@webex/spark-widget-base/src/enhancers/withBrowserGlobals.js
+++ b/packages/node_modules/@webex/spark-widget-base/src/enhancers/withBrowserGlobals.js
@@ -30,7 +30,7 @@ function BrowserWidget(el) {
     const removed = ReactDOM.unmountComponentAtNode(el);
 
     // Remove from widgetStore
-    Reflect.deleteProperty(window.ciscospark.widgetStore, id);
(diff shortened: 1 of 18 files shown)


packages/node_modules/@webex/spark-widget-base/src/enhancers/withBrowserGlobals.js
```diff
@@ -166,11 +170,11 @@ function setupSparkWidgetNamespace() {
  */
 export default function withBrowserGlobals({name}) {
   // Inject widget into browser globals
-  setupSparkWidgetNamespace();
+  setupWidgetNamespace();
   const fnName = `${name}Widget`;
 
   return (BaseComponent) => {
-    function renderSparkWidget(options) {
```
InteractiveTimmy: Not sure if you wanted to include `Webex` here, but looks good otherwise.

adamweeks: I figured removing the product name from the function name would make it a bit more future proof. It doesn't need to be specific since there's only one type of widget in this codebase.

PR #1047, opened 2020-01-17
fix(widget-recents): fix 'leave' issues

mwegman:
- only remove space when the current user leaves
- don't fetch an unknown space if current user left it


mwegman:
Missing tests

PR #1048, opened 2020-01-21
update to Babel v7

taymoork2:


packages/node_modules/@ciscospark/react-component-activity-item/src/index.js
```diff
@@ -42,6 +44,7 @@ const propTypes = {
 
 const defaultProps = {
   isAdditional: false,
+  // eslint-disable-next-line react/default-props-match-prop-types
```
adamweeks: You could delete the `sdkInstance` since it is a required prop type. No need to add the eslint override.

packages/node_modules/@ciscospark/react-component-activity-share-thumbnail/src/index.js
```diff
@@ -17,6 +17,8 @@ const propTypes = {
   file: PropTypes.shape({
     displayName: PropTypes.string,
     fileSize: PropTypes.number,
+    image: PropTypes.shape(),
```
adamweeks: This should be object.
```suggestion
    image: PropTypes.object(),
```

packages/node_modules/@ciscospark/react-component-file-staging-area/src/index.js
```diff
@@ -29,6 +29,7 @@ function FileStagingArea({
 
   const fileChips = [];
 
+  // eslint-disable-next-line react/prop-types
```
adamweeks: Are the prop types wrong on this? Code seems to indicate that `files` is an array.

packages/node_modules/@webex/react-component-activity-item/src/index.js
```diff
@@ -42,6 +44,7 @@ const propTypes = {
 
 const defaultProps = {
   isAdditional: false,
+  // eslint-disable-next-line react/default-props-match-prop-types
   sdkInstance: {}
```
adamweeks: Another `sdkInstance` that can be removed because it is required.

packages/node_modules/@webex/react-component-activity-share-thumbnail/src/index.js
```diff
@@ -17,6 +17,8 @@ const propTypes = {
   file: PropTypes.shape({
     displayName: PropTypes.string,
     fileSize: PropTypes.number,
+    image: PropTypes.shape(),
```
adamweeks:
```suggestion
    image: PropTypes.object(),
```

adamweeks:
Good first pass, nice to see all the tests passing! Comments are mostly around the react prop changes.

package.json
```diff
@@ -112,7 +113,7 @@
     "react-mentions": "^1.2.1",
     "react-redux": "^5.0.1",
     "react-syntax-highlighter": "^6.0.2",
-    "react-virtualized": "^9.18.5",
+    "react-virtualized": "^9.21.0",
```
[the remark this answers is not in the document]
taymoork2: I was running into failures on jest with react-virtualized's `scrollbarSize` not being found. I thought updating it to latest would fix it, however the problem persisted.
Looking into it, there was a version mismatch of a dependency (`dom-helpers`) with `react-virtualized` and `momentum-ui` where momentum-ui was still using the older version of `dom-helpers`.
_"Sticking"_ `react-virtualized` to `9.21.0` keeps the `dom-helpers` dependency inline with the same version that `momentum-ui` uses.

https://github.com/bvaughn/react-virtualized/issues/1458#issuecomment-552925668

packages/node_modules/@ciscospark/react-component-activity-share-thumbnail/src/index.js
taymoork2: I think I originally made this a shape since, I believe, there are two properties associated with image, `height` & `width`, and added them into the proptype, however wasnt sure if height & width were being passed as strings or numbers

packages/node_modules/@ciscospark/react-component-file-staging-area/src/index.js
taymoork2: For some reason it's indicating that `forEach` is not a builtin, but a prop that needs to be validated.
Not entirely sure what the solution here was besides just ignoring it

package.json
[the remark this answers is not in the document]
taymoork2: Retried without bumping, problem isnt showing up anymore, so reverting back 9.18.5

packages/node_modules/@ciscospark/react-component-file-staging-area/src/index.js
adamweeks: Gotcha, yea let's ignore it. Looking at the tests it is an immutable map that is expected.

packages/node_modules/@ciscospark/react-component-activity-share-thumbnail/src/index.js
adamweeks: Right, but a shape prop type indicates that you will provide an actual shape. `shape()` doesn't tell us anything about the property, so let's just use the ambiguous `.object`

PR #1053, opened 2020-01-30
Merge pull request #1053 from taymoork2/npm-ci

taymoork2:
On CircleCI we currently restore cache before running npm i on the install jobs. This has caused a weird issue where sometimes post merge npm install has caused package updates to not be reflected/installed correctly.
Moving to not restore cache and then using npm ci and then saving the cache should stop that from happening

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,10 +36,9 @@ jobs:
     executor: main-executor
     steps:
       - checkout
-      - restore_node_modules
       - run:
           name: 'Install dependencies'
-          command: npm i
+          command: npm ci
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -36,10 +36,9 @@ jobs:
     executor: main-executor
     steps:
       - checkout
-      - restore_node_modules
       - run:
           name: 'Install dependencies'
-          command: npm i
+          command: npm ci
```
InteractiveTimmy: Good change, didn't know npm had this feature.

PR #1057, opened 2020-02-04
Remove all `@ciscospark` packages

taymoork2:


adamweeks:
No need to add the deprecate.sh script, otherwise looks good.

PR #1072, opened 2020-03-25
Add pagerduty to tap test job

taymoork2:


adamweeks:
Do we need to add `PAGERDUTY_KEY` to the env vars on circle?

taymoork2:
> Do we need to add `PAGERDUTY_KEY` to the env vars on circle?

Correct

.circleci/config.yml
```diff
@@ -350,6 +350,17 @@ jobs:
       - store_artifacts:
           path: /home/circleci/.npm/_logs/
           destination: npm-logs
+      - run:
+          name: 'Create Incident on PagerDuty'
+          when: on_fail
+          command: |
+            curl -X POST --header 'Content-Type: application/json' -d '{
```
InteractiveTimmy: Could / Have we set up a trial run of this CURL request? Would like to confirm before we proceed with merging.

.circleci/config.yml
```diff
@@ -350,6 +350,17 @@ jobs:
...
+              "service_key": "${PAGERDUTY_KEY}",
+              "event_type": "trigger",
+              "description": "FAILURE -- React Widgets Tap Test -- ${CIRCLE_BUILD_URL}",
+              "client": "CircleCI - React Widgets",
+              "client_url": "${CIRCLE_BUILD_URL}"
+            }' https://events.pagerduty.com/generic/2010-04-15/create_event.json
```
lalli-flores: Is there a reason why you didn't use Pager Duty [Event API v2](https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2)?

.circleci/config.yml
```diff
@@ -350,6 +350,17 @@ jobs:
...
           path: /home/circleci/.npm/_logs/
           destination: npm-logs
+      - run:
+          name: 'Create Incident on PagerDuty'
+          when: on_fail
+          command: |
+            curl -X POST --header 'Content-Type: application/json' -d '{
+              "service_key": "${PAGERDUTY_KEY}",
+              "event_type": "trigger",
+              "description": "FAILURE -- React Widgets Tap Test -- ${CIRCLE_BUILD_URL}",
```
lalli-flores: Should probably add the build URL as part of the details. Having it as part of the summary would make it too crowded/long.

```suggestion
              "description": "FAILURE: React Widgets TAP Test",
```

.circleci/config.yml
taymoork2: v2 requires a "from" property which is essentially an email, this just uses the key as authentication which i think its just less setup and cleaner

.circleci/config.yml
taymoork2: Yeah that makes sense

.circleci/config.yml
lalli-flores: am I missing something? Where is the `from` parameter?
Screen Shot 2020-03-27 at 10 12 30 AM

PR #1079, opened 2020-04-14
feat(webex): set audio and video dynamically to calls in widgets

brunogenaro:
I noticed we have unit test only for the functions that dispatch an action in `redux-module-media/src/actions.js` so I didn't add unit test for this new function.  This is just a helper function to automatically set `audio` and `video` constraints before calling `placeCall`.

I'm having a hard time to test it locally since it needs https to have access to the `navigator.mediaDevices` object and its methods. 

Is it possible to deploy it to Netlify and test it before we merged it?

### DIFF
--- a/packages/node_modules/@webex/redux-module-media/src/actions.js
+++ b/packages/node_modules/@webex/redux-module-media/src/actions.js
@@ -218,6 +218,31 @@ export function listenForCalls(sparkInstance) {
 }
 
 /**
+ * Checks if the machine has at least one audio or video device
+ * @returns {Promise.<Object>} {
+ *    audio: true/false,
+ *    video: true/false
+ *}
+ */
+function getSupportedDevice() {
+  return Promise.resolve().then(() => {
+    if (!navigator.mediaDevices || navigator.mediaDevices.enumerateDevices === undefined) {
+      return {
+        audio: false,
+        video: false
+      };
+    }
+
+    // Reference for device.kind: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices
+    return navigator.mediaDevices.enumerateDevices()
+      .then((devices) => ({
+        audio: devices.filter((device) => device.kind === 'audioinput').length > 0,
+        video: devices.filter((device) => device.kind === 'videoinput').length > 0
+      }));
+  });
+}
+
+/**
  * Call a user with an email address or userId
  *
  * @export
@@ -226,30 +251,32 @@ export function listenForCalls(sparkInstance) {
  * @param {Object} data.options
  * @returns {Promise}
  */
-export function placeCall(sparkInstance, {
-  destination,
-  options = {audio: true, video: true}
-}) {
+export function placeCall(sparkInstance, {destination, options}) {
   return (dispatch) => {
     let call, id;
 
-    try {
-      call = sparkInstance.phone.dial(
-        destination,
-        Object.assign(
-          options,
-          {
-            offerOptions: {offerToReceiveVideo: true, offerToReceiveAudio: true}
-          }
-        )
-      );
-      ({call, id} = processCall({dispatch, call}));
-    }
-    catch (e) {
-      // Ignore, will be caught on call.on('error')
-    }
+    return getSupportedDevice()
+      .then(({audio, video}) => {
+        try {
+          call = sparkInstance.phone.dial(
+            destination,
+            Object.assign(
+              options,
+              {
+                offerOptions: {offerToReceiveVideo: true, offerToReceiveAudio: true},
+                audio,
+                video
+              }
+            )
+          );
+          ({call, id} = processCall({dispatch, call}));
+        }
+        catch (e) {
+          // Ignore, will be caught on call.on('error')
+        }
 
-    return Promise.resolve({call, id});
+        return {call, id};
+      });
   };
 }
 


packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -217,6 +217,31 @@ export function listenForCalls(sparkInstance) {
   };
 }
 
+/**
+ * Checks if the machine has at least one audio or video device
+ * @returns {Object} {
```
adamweeks, line 222:
```suggestion
 * @returns {Promise.<Object>} {
```

packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -226,28 +251,32 @@ export function listenForCalls(sparkInstance) {
...
-            offerOptions: {offerToReceiveVideo: true, offerToReceiveAudio: true}
-          }
-        )
-      );
-      ({call, id} = processCall({dispatch, call}));
-    }
-    catch (e) {
-      // Ignore, will be caught on call.on('error')
-    }
+    getSupportedDevice()
```
adamweeks, line 258: Return this promise as part of your dispatch:
```suggestion
    return getSupportedDevice()
```

packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -226,28 +251,32 @@ export function listenForCalls(sparkInstance) {
...
+          ({call, id} = processCall({dispatch, call}));
+        }
+        catch (e) {
+          // Ignore, will be caught on call.on('error')
+        }
+
+        return Promise.resolve({call, id});
+      });
 
     return Promise.resolve({call, id});
```
adamweeks, line 281: This line can be removed, it shouldn't be reached once you `return getSupportedDevice`

packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -226,28 +251,32 @@ export function listenForCalls(sparkInstance) {
...
+              }
+            )
+          );
+          ({call, id} = processCall({dispatch, call}));
+        }
+        catch (e) {
+          // Ignore, will be caught on call.on('error')
+        }
+
+        return Promise.resolve({call, id});
```
adamweeks, line 278: Since this is nested in a `then` you can just do this:
```suggestion
        return {call, id};
```

packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -217,6 +217,31 @@ export function listenForCalls(sparkInstance) {
...
+/**
+ * Checks if the machine has at least one audio or video device
+ * @returns {Object} {
+ *    audio: true/false,
+ *    video: true/false
+ *}
+ */
+export function getSupportedDevice() {
+  return Promise.resolve().then(() => {
+    if (!navigator.mediaDevices || navigator.mediaDevices.enumerateDevices === undefined) {
```
adamweeks, line 229: If we can't enumerate the devices, should we default them to `false` instead? Maybe @taymoork2 can provide some more insight.

brunogenaro: any insight on this @taymoork2  ?

taymoork2: probably best to default to false than take the chance of setting it to true

taymoork2: however we probably will start seeing issues with IE11 since enumerateDevices isnt available in ie11

https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices

brunogenaro: But currently do we support IE11 at all?

taymoork2: I don't think that IE11 has a ton, if not very little, webrtc support
also this is implemented on the SDK

packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -217,6 +217,31 @@ export function listenForCalls(sparkInstance) {
...
 }
 
+/**
+ * Checks if the machine has at least one audio or video device
+ * @returns {Object} {
+ *    audio: true/false,
+ *    video: true/false
+ *}
+ */
+export function getSupportedDevice() {
```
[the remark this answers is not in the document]
brunogenaro: done!

packages/node_modules/@webex/redux-module-media/src/actions.js
brunogenaro: done!

packages/node_modules/@webex/redux-module-media/src/actions.js
brunogenaro: done!

packages/node_modules/@webex/redux-module-media/src/actions.js
brunogenaro: done!

packages/node_modules/@webex/redux-module-media/src/actions.js
brunogenaro: done!

packages/node_modules/@webex/redux-module-media/src/actions.js
brunogenaro: done!

brunogenaro:
Thanks for the feedback! Like I mentioned it was kinda hard because I wasn't actually able to test it locally.

packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -217,6 +217,31 @@ export function listenForCalls(sparkInstance) {
...
+ * Checks if the machine has at least one audio or video device
+ * @returns {Promise.<Object>} {
+ *    audio: true/false,
+ *    video: true/false
+ *}
+ */
+function getSupportedDevice() {
+  return Promise.resolve().then(() => {
+    if (!navigator.mediaDevices || navigator.mediaDevices.enumerateDevices === undefined) {
+      return {
```
lalli-flores, line 230: Are you returning two different things? this is an object and the other return statement is returning a promise

brunogenaro: @lalli-flores I noticed that as well but I'm having difficult to test it locally to make sure it works. Also, that's the same approach used in the `webex-js-sdk` repo: https://github.com/webex/webex-js-sdk/blob/ba7012680e9b3a8f07f5a662dfdced485772ab43/packages/node_modules/@webex/plugin-meetings/src/media/index.js#L372

I'm not sure if that is something that is wrong there as well.

adamweeks: It is returning from within a promise on line 228, so technically it is returning a promise as well.

lalli-flores: I got it, [from the spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve), if the return value is a `thenable` the returned promise will follow the `then`.

brunogenaro: It's all good then @adamweeks?

PR #1089, opened 2020-05-05
fix(redux-module-spaces): add code to error message

InteractiveTimmy:
The scope of the changes in this pull request is adding explicit error details to the already implemented fix for [this case](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-132963). This will allow users to see an explicit messaged in the view based on the data retrieved from the JS SDK.

### DIFF
--- a/packages/node_modules/@webex/redux-module-spaces/src/__snapshots__/actions.test.js.snap
+++ b/packages/node_modules/@webex/redux-module-spaces/src/__snapshots__/actions.test.js.snap
@@ -506,7 +506,7 @@ Array [
     "payload": Object {
       "error": Object {
         "code": "Error",
-        "displaySubtitle": "Unable to load spaces. Please try again later.",
+        "displaySubtitle": "Unable to load spaces. Please try again later. [Error]",
         "displayTitle": "Something's not right",
         "id": "redux-module-spaces-load",
         "temporary": false,
@@ -736,7 +736,7 @@ Array [
     "payload": Object {
       "error": Object {
         "code": "Error",
-        "displaySubtitle": "Unable to load spaces. Please try again later.",
+        "displaySubtitle": "Unable to load spaces. Please try again later. [Error]",
         "displayTitle": "Something's not right",
         "id": "redux-module-spaces-load",
         "temporary": false,
--- a/packages/node_modules/@webex/redux-module-spaces/src/actions.js
+++ b/packages/node_modules/@webex/redux-module-spaces/src/actions.js
@@ -117,14 +117,14 @@ function storeInitialSpace(id) {
 /**
  * Adds a rate limit error to the errors module
  * @param {function} dispatch
- * @param {String} code
+ * @param {String} name
  */
-function addLoadError(dispatch, code) {
+function addLoadError(dispatch, name) {
   dispatch(addError({
-    code,
+    code: name,
     id: 'redux-module-spaces-load',
     displayTitle: 'Something\'s not right',
-    displaySubtitle: 'Unable to load spaces. Please try again later.',
+    displaySubtitle: `Unable to load spaces. Please try again later. [${name}]`,
     temporary: false
   }));
 }


packages/node_modules/@webex/redux-module-spaces/src/__snapshots__/actions.test.js.snap
```diff
@@ -506,7 +506,7 @@ Array [
     "payload": Object {
       "error": Object {
         "code": "Error",
-        "displaySubtitle": "Unable to load spaces. Please try again later.",
+        "displaySubtitle": "Unable to load spaces. Please try again later. [Error]",
```
lalli-flores, line 509: What's the value of adding `[Error]` at the end of the message?

packages/node_modules/@webex/redux-module-spaces/src/actions.js
```diff
@@ -124,7 +124,7 @@ function addLoadError(dispatch, code) {
     code,
     id: 'redux-module-spaces-load',
     displayTitle: 'Something\'s not right',
-    displaySubtitle: 'Unable to load spaces. Please try again later.',
+    displaySubtitle: `Unable to load spaces. Please try again later. [${code}]`,
```
lalli-flores, line 127: Is this adding the error code from the HTTP call?

packages/node_modules/@webex/redux-module-spaces/src/__snapshots__/actions.test.js.snap
InteractiveTimmy: This allows users to better identify the cause of a failed load. The only presentation portions of the redux store object that are presented to the user are `displayTitle` and `displaySubtitle`, which aren't dynamic or helpful unless they present some form of error details.

From the JIRA's scope:
>Utilize the redux error store to add an error that will be displayed by the root widget.

packages/node_modules/@webex/redux-module-spaces/src/actions.js
InteractiveTimmy: This adds the mutated code from the HTTP call, the JS SDK converts the code into a string, and it is provided in this method as the `code` parameter.

packages/node_modules/@webex/redux-module-spaces/src/__snapshots__/actions.test.js.snap
adamweeks: @lalli-flores  It won't actually display "Error", it will display the error that is being passed to it. Since these are snapshots, it is just what we are passing. Perhaps the test could be updated a bit to have a more realistic error @InteractiveTimmy?

packages/node_modules/@webex/redux-module-spaces/src/actions.js
adamweeks: It looks like all the calls to this function are doing:
```js
.catch((err) => {
        addLoadError(dispatch, err.name);

        throw (err);
      });
```

We should change the param name from "code" to "name"? Not really in scope of this ticket, but could be a quick win.

packages/node_modules/@webex/redux-module-spaces/src/__snapshots__/actions.test.js.snap
InteractiveTimmy: These were auto-updated via jest, I can manually improve them though. Let me know if you'd like me to proceed with doing this.

packages/node_modules/@webex/redux-module-spaces/src/actions.js
InteractiveTimmy: The name can also exist as a code if it isn't an error code that is mapped by the JS SDK. So, if we receive something like `999`, it will be `999` in the brackets, as it doesn't map to a known name in the SDK, see the status code interceptor for more details.

packages/node_modules/@webex/redux-module-spaces/src/__snapshots__/actions.test.js.snap
lalli-flores: OMG! My bad, totally didn't realize this where snapshots 😂🤦🏻‍♀️

packages/node_modules/@webex/redux-module-spaces/src/actions.js
lalli-flores: yeah, calling it `code` is misleading

InteractiveTimmy: update this snippet.

packages/node_modules/@webex/redux-module-spaces/src/actions.js
```diff
@@ -117,14 +117,14 @@ function storeInitialSpace(id) {
...
  */
-function addLoadError(dispatch, code) {
+function addLoadError(dispatch, name) {
   dispatch(addError({
-    code,
+    code: name,
     id: 'redux-module-spaces-load',
     displayTitle: 'Something\'s not right',
-    displaySubtitle: 'Unable to load spaces. Please try again later.',
+    displaySubtitle: `Unable to load spaces. Please try again later. [${name}]`,
```
lalli-flores, line 127: Should the error be on a different line? or have a nicer message of the error?

InteractiveTimmy: Potentially, this means we'll be injecting formatting into a string that did not have formatting originally, resulting in a coupling of business logic with presentation logic.

We could create a larger effort to fix the presentation of all error actions within the redux store.

PR #1095, opened 2020-06-12
Ability to disable dropzone when files is set to false

taymoork2:


packages/node_modules/@webex/widget-message/src/container.js
```diff
@@ -741,7 +741,8 @@ export class MessageWidget extends Component {
       currentUser,
       muteNotifications,
       widgetMessage,
-      features
+      features,
+      spaceActivities
```
mwegman, line 745: Doesn't a reducer have to be added to account for this prop? Who is passing this prop in?

taymoork2: It's a container, so it's using redux/compose/etc to grab stuff from the store and then injecting it.
This a big file and injects a lot into it's state and props (I'm guessing to drill down), so where it's actually grabbing from ¯\_(ツ)_/¯, but while digging through react dev tools showed that it was prop in this MessageWidget container.

InteractiveTimmy: It seems as though `spaceActivities` is coming `widget-space`. I'm not sure how this data gets from that component to this component, but if it successfully passes tests and works in a live run, it has my approval. Let me know if this is the case, I'll be standing by for the tests to pass on CI.

packages/node_modules/@webex/widget-message/src/container.js
```diff
@@ -753,6 +755,7 @@ export class MessageWidget extends Component {
         <Dropzone
```
InteractiveTimmy, line 755: Dropzone is from `react-drop-zone`, so no additional unit tests are needed for this use case.

PR #1105, opened 2020-07-21
Sanitize input before displaying

adamweeks:
This PR adds the ability to sanitize any text sent to the DOM from the activity post component.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-148479

### DIFF
--- a/packages/node_modules/@webex/react-component-utils/src/activity.js
+++ b/packages/node_modules/@webex/react-component-utils/src/activity.js
@@ -1,5 +1,7 @@
 import uuid from 'uuid';
+
 import {base64} from '@webex/common';
+import {escapeSync} from '@webex/helper-html';
 
 import {isImage} from './files';
 
@@ -10,9 +12,10 @@ import {isImage} from './files';
  * @param {object} conversation
  * @param {object} activityObject
  * @param {any} actor
+ * @param {boolean} [sanitize = false]
  * @returns {object}
  */
-export function constructActivity(conversation, activityObject, actor) {
+export function constructActivity(conversation, activityObject, actor, sanitize = false) {
   const clientTempId = `sdk-widget-${uuid.v4()}`;
 
   return {
@@ -25,7 +28,10 @@ export function constructActivity(conversation, activityObject, actor) {
     clientTempId,
     id: clientTempId,
     // Minimum properties needed by API
-    object: activityObject,
+    object: sanitize ? {
+      ...activityObject,
+      content: activityObject.content && escapeSync(activityObject.content)
+    } : activityObject,
     target: {
       id: conversation.id,
       objectType: 'conversation'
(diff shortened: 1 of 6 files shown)


packages/node_modules/@webex/react-component-utils/src/activity.js
```diff
@@ -10,11 +12,23 @@ import {isImage} from './files';
...
   const clientTempId = `sdk-widget-${uuid.v4()}`;
 
+  const modifiedActivityObject = Object.assign({}, activityObject);
+
+  if (sanitize) {
+    if (modifiedActivityObject.displayName) {
+      modifiedActivityObject.displayName = DOMPurify.sanitize(activityObject.displayName);
+    }
+    if (modifiedActivityObject.content) {
+      modifiedActivityObject.content = DOMPurify.sanitize(activityObject.content);
```
mwegman, line 28: Instead of using `DOMPurify` here, you should use `escape` from `@webex/helper-html` when the object is being created.

taymoork2: ...something something i also expect tests 😈

packages/node_modules/@webex/react-component-utils/src/activity.js
```diff
@@ -10,11 +12,23 @@ import {isImage} from './files';
...
  */
-export function constructActivity(conversation, activityObject, actor) {
+export function constructActivity(conversation, activityObject, actor, sanitize = false) {
   const clientTempId = `sdk-widget-${uuid.v4()}`;
 
+  const modifiedActivityObject = Object.assign({}, activityObject);
+
+  if (sanitize) {
+    if (modifiedActivityObject.displayName) {
+      modifiedActivityObject.displayName = escapeSync(activityObject.displayName);
```
mwegman, line 25: Don't do it for `displayName`

adamweeks: Why? It will display one or the other if there is no content

adamweeks: We will render either https://github.com/adamweeks/react-ciscospark/blob/6a3a8a037cd4c7004305aff69405c55cec03953c/packages/node_modules/@webex/react-component-activity-text/src/index.js#L29 so we need to sanitize both before sending to dom

taymoork2: The other spot where sanitization happens in the widgets (done by the adaptive cards team), they also modify the `displayName`
https://github.com/webex/react-widgets/blob/master/packages/node_modules/@webex/redux-module-activity/src/helpers.js#L101

mwegman: We should not be displaying `displayName` like that. That is a bug.
If `content` is not there, you use `displayName` but you don't set it like it is HTML. One should display it like a string of characters.

adamweeks: @mwegman I added another commit that fixes the text render component to not html render `displayName` and have removed the sanitization of that property.

packages/node_modules/@webex/react-component-utils/src/activity.js
```diff
@@ -25,7 +36,7 @@ export function constructActivity(conversation, activityObject, actor) {
     clientTempId,
     id: clientTempId,
     // Minimum properties needed by API
-    object: activityObject,
+    object: modifiedActivityObject,
```
taymoork2, line 39:
```suggestion
  object: sanitize ?
    {
      ...activityObject,
      content: escapeSync(activityObject.content)
    } :
    activityObject;
```

adamweeks: Done

Issue #1091, opened 2020-05-15
Widgets not working: babelHelpers is not defined


saltenhub:
# Bug report
I am not able to get any widget working with a current version of create-react-app (react-scripts v3.4.1) - the build finished fine, but when I try to access the page I am running into a reference error:
```
ReferenceError: babelHelpers is not defined
```

![Error message](https://user-images.githubusercontent.com/52566130/82073664-d98a2380-96d9-11ea-9d17-0a6ac1b54c67.png)

**Steps to reproduce:**
- Create a react app with `yarn create react-app` (react-scripts current version - v3.4.1)
- Add dependency for any widget, e.g. `yarn add @webex/widget-space` (v0.2.17)
- Import widget for use via `import SpaceWidget from @webex/widget-space`
- Add @formatjs/intl-pluralrules because of #1090 `yarn add @formatjs/intl-pluralrules`
- `yarn start` to start the development server
- Go to `http://localhost:3000`

**Expected behavior**:
The widget working as described in the documentation.

machner:
Same here - without importing the @webex/widget-space it works fine. We can't get the widget to run even in minimal react app.
![ebabff6c-3f53-4afd-80d9-a0c891150436](https://user-images.githubusercontent.com/40216144/82092225-27faea80-96f9-11ea-9839-9765381036ec.PNG)

nhogge:
I found a workaround...

install babel cli
`npm i @babel/cli -D`

generate the babelHelpers js
`npx babel-external-helpers > babel-helpers.js`

include in your project first. I am using webpack to did the following:
`entry: [
    './src/babel-helpers.js',
    './src/index.js'
  ],`

Haas-webquake:
Thanks for the support !

I have done the first two points, but now I am not sure where in the webpack I should insert the babel-helpers.js.

I found the webpack.config.js in the folder react-scripts/config/. I think this is the right file. I'm posting here a screenshot of my attempt to integrate your code. May I ask you for your support again.

Thanks in advance, best regards

![image](https://user-images.githubusercontent.com/65902178/82812141-9e82af80-9e92-11ea-86d0-dce8dd9342f3.png)

jdaly13:
This should be fixed with the latest PR that has been merged into master
#1110

This would be release 0.2.115

Issue #1090, opened 2020-05-15
Missing dependency @formatjs/intl-pluralrules


saltenhub:
# Bug report

It seems to me that the dependency for `@formatjs/intl-pluralrules` is missing from `@webex/webex-widget-base`:
```
Failed to compile.

./node_modules/@webex/webex-widget-base/es/index.js
Module not found: Can't resolve '@formatjs/intl-pluralrules/dist/locale-data' in '/home/saltenhub/webex-test/node_modules/@webex/webex-widget-base/es'
```

**Steps to reproduce:**
- Create a react app with `yarn create react-app` (react-scripts current version - v3.4.1)
- Add dependency for any widget, e.g. `yarn add @webex/widget-space` (v0.2.17)
- Import widget for use via `import SpaceWidget from @webex/widget-space`
- `yarn start` to start the development server

**Expected behavior:**
`@formatjs/intl-pluralrules` gets pulled correctly as a dependency on adding the `@webex/widget-space` and does not need to be added to `package.json` manually.

**Interim solution:**
`yarn add @formatjs/intl-pluralrules` solves the issue for me.

oborys:
I have the same issue

mchandran:
Receiving error:- Module not found: Can't resolve '@formatjs/intl-pluralrules/dist/locale-data'
npm add @formatjs/intl-pluralrules  - Didn't solve the issue.

jdaly13:
This should be fixed with the latest PR that has been merged into master
https://github.com/webex/react-widgets/pull/1110

This would be release 0.2.115

PR #1110, opened 2020-08-04
Dependecy and configuration update

jdaly13:
As mentioned in the jira task [165188](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-165188?focusedCommentId=2281841&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2281841) 
This PR is a bug fix that will include the proper dependencies regardless of which dynamic formatjs plugin is required by withIntl.js file
This PR also fixes the babelHelpers issue mentioned in this github [issue](https://github.com/webex/react-widgets/issues/1091)
So this PR should fix these two github issues
https://github.com/webex/react-widgets/issues/1090
and
https://github.com/webex/react-widgets/issues/1091


jdaly13:
See these two issues regarding babelHelpers as well for context
https://github.com/babel/babel/issues/9366
https://github.com/rollup/rollup/issues/2665

https://babeljs.io/docs/en/6.26.3/babel-plugin-external-helpers#detail

packages/node_modules/@webex/webex-widget-base/package.json
```diff
@@ -7,6 +7,10 @@
   "keywords": [],
   "author": "devsupport@webex.com",
   "license": "MIT",
+  "dependencies": {
```
adamweeks, line 10: Unfortunately, these get overwritten by the build system. If you run `npm run build package.json` you'll see that it no longer has this dependency.

Whatever solution we come up with will need to work with or along side of the [deps.js](https://github.com/adamweeks/react-widgets/blob/e9f321b96216fee09aa0905ab03d29300a211b31/scripts/utils/deps.js#L19) script.

.gitignore
```diff
@@ -7,6 +7,7 @@
 npm-debug*
 packages/**/es/*
 packages/**/cjs/*
+packages/**/node_modules/*
```
adamweeks, line 10: Let's scope this change in its own commit.

packages/node_modules/@webex/webex-widget-base/package.json
jdaly13: I updated the deps.js script to look for existing dependencies first and if not just create a simple object

jdaly13: that should fix the issue

jdaly13:
Hey @adamweeks let me know what you want to do - I have 3 commits.  Do you want me to revert commit 1?  I can possibly change the first commit message but I don't think I can edit that commit (meaning removing the gitignore changes) unless I revert commit.  I think the easiest thing would be to possibly revert the first commit and then do two more commits - one for the original changes with updated commit message and then another commit with the gitignore changes.  I added those gitignore changes because when you add a dependency and do es build it will add node_modules to the package and you have hundreds of new files that would be committed.

adamweeks:
@jdaly13 you can rebase and split the first commit into two in order to get the `gitcommit` changes into their own commit. See this SO question for steps on how to do it: https://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits

Also, the last commit message should be `chore(tooling)` for the build process script change.

jdaly13:
@adamweeks I have split up those commits while amending the commit messages as well, this should satisfy requirements.

PR #1111, opened 2020-08-11
Set limit on attachments

jdaly13:
This PR is to fix [160675](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-160675).  When a single attachment is larger than 150mb on chrome or firefox it causes the browser to crash when the files are added, This PR fixes that by adding a threshold property that will be true if file size is over 150mb.  Right now the only cue to user is an alert that tells them one of the attached files is too large.

### DIFF
--- a/packages/node_modules/@webex/react-component-utils/src/constants.js
+++ b/packages/node_modules/@webex/react-component-utils/src/constants.js
@@ -86,6 +86,8 @@ export const TEXT_INPUT_ELEMENT = 'Input.Text';
 export const TOGGLE_INPUT_ELEMENT = 'Input.Toggle';
 export const CHOICE_SET_INPUT_ELEMENT = 'Input.ChoiceSet';
 
+export const MAXIMUM_BYTES = 150000000; // 150 MB threshold limit
+
 export const ADAPTIVE_CARD_HOST_CONFIG = {
   hostCapabilities: {
     capabilities: null
--- a/packages/node_modules/@webex/react-component-utils/src/constants.js
+++ b/packages/node_modules/@webex/react-component-utils/src/constants.js
@@ -86,7 +86,7 @@ export const TEXT_INPUT_ELEMENT = 'Input.Text';
 export const TOGGLE_INPUT_ELEMENT = 'Input.Toggle';
 export const CHOICE_SET_INPUT_ELEMENT = 'Input.ChoiceSet';
 
-export const MAXIMUM_BYTES = 150000000; // 150 MB threshold limit
+export const FILE_ATTACHMENT_MAX_SIZE = 150000000; // 150 MB threshold limit
 
 export const ADAPTIVE_CARD_HOST_CONFIG = {
   hostCapabilities: {
--- a/packages/node_modules/@webex/react-component-utils/src/files.js
+++ b/packages/node_modules/@webex/react-component-utils/src/files.js
@@ -1,6 +1,5 @@
 import uuid from 'uuid';
 
-
 const FILE_TYPES = {
   'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet': 'spreadsheet',
   'application/pdf': 'pdf',
--- a/packages/node_modules/@webex/react-component-utils/src/files.js
+++ b/packages/node_modules/@webex/react-component-utils/src/files.js
@@ -1,6 +1,5 @@
 import uuid from 'uuid';
 
-import {FILE_ATTACHMENT_MAX_SIZE} from './constants';
 
 const FILE_TYPES = {
   'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet': 'spreadsheet',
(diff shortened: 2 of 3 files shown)


packages/node_modules/@webex/react-container-message-composer/src/container.js
```diff
@@ -177,8 +177,15 @@ export class MessageComposer extends Component {
       } = props;
 
       const files = constructFiles(e.target.files);
+      const overThreshold = files.some((file) => file.overThreshold);
 
-      props.addFiles(conversation, activity, files, sparkInstance);
+      if (overThreshold) {
+        // eslint-disable-next-line no-alert
+        alert('One of your files is too large to attach');
```
adamweeks, line 184: Take a look at the `addError` function to display an error across widgets:

```js
import {addError} from '@webex/redux-module-errors';
```

packages/node_modules/@webex/react-component-utils/src/constants.js
```diff
@@ -86,6 +86,8 @@ export const TEXT_INPUT_ELEMENT = 'Input.Text';
 export const TOGGLE_INPUT_ELEMENT = 'Input.Toggle';
 export const CHOICE_SET_INPUT_ELEMENT = 'Input.ChoiceSet';
 
+export const MAXIMUM_BYTES = 150000000; // 150 MB threshold limit
```
lalli-flores, line 89: Can you make this name more specific?
```suggestion
export const FILE_ATTACHMENT_MAX_SIZE = 150000000; // 150 MB threshold limit
```

packages/node_modules/@webex/react-container-message-composer/src/container.js
```diff
@@ -177,8 +180,21 @@ export class MessageComposer extends Component {
...
-
-      props.addFiles(conversation, activity, files, sparkInstance);
+      const overThreshold = files.some((file) => file.overThreshold);
+
+      if (overThreshold) {
+        props.addError({
+          id: 'file-attachment-threshold',
+          actionTitle: 'Dismiss',
+          onAction: () => props.removeError('file-attachment-threshold'),
+          displayTitle: 'What if I told you',
```
lalli-flores, line 190: Can you add a screenshot on the PR description as to how this alert looks like?

jdaly13: Screenshot 2020-08-13 at 3 33 49 PM
Screenshot 2020-08-13 at 3 34 02 PM

jdaly13: when you click dismiss it goes back to normal state

packages/node_modules/@webex/react-component-utils/src/files.js
```diff
@@ -51,7 +53,8 @@ export function constructFile(file) {
     displayName: file.name,
     fileSize: file.size,
     fileSizePretty: bytesToSize(file.size),
-    mimeType: file.type
+    mimeType: file.type,
+    overThreshold: file.size >= FILE_ATTACHMENT_MAX_SIZE
```
lalli-flores, line 57: Sorry I didn't comment on this earlier, I just understood what you're doing.
Since the file size is already passed as part of the file object, I don't we should add this as a property, as it can be computed dynamically in the component.
Hence my suggestion is to just do the check in the component.

packages/node_modules/@webex/react-container-message-composer/src/container.js
```diff
@@ -177,8 +180,21 @@ export class MessageComposer extends Component {
       } = props;
 
       const files = constructFiles(e.target.files);
-
-      props.addFiles(conversation, activity, files, sparkInstance);
+      const overThreshold = files.some((file) => file.overThreshold);
```
lalli-flores, line 183: Following up on my comment to not store the check as a property:
```suggestion
      const maxSizeExceeded = files.some((file) => file.size >= FILE_ATTACHMENT_MAX_SIZE);
```

packages/node_modules/@webex/react-container-message-composer/src/container.js
lalli-flores: Thanks for adding the screenshot 👍🏼
Here's my wording suggestion:
```suggestion
          displayTitle: 'Too big for the web!',
```

packages/node_modules/@webex/react-container-message-composer/src/container.js
```diff
@@ -177,8 +180,21 @@ export class MessageComposer extends Component {
...
+
+      if (overThreshold) {
+        props.addError({
+          id: 'file-attachment-threshold',
+          actionTitle: 'Dismiss',
+          onAction: () => props.removeError('file-attachment-threshold'),
+          displayTitle: 'What if I told you',
+          displaySubtitle: 'A file you are attempting to attach is too large and will crash the browser',
```
lalli-flores, line 191: I think we should let the user know exactly what they did wrong
```suggestion
          displaySubtitle: Only attachments smaller than 150MB are allowed' ,
```

jdaly13:
All your concerns have been addressed - should be good to go.

lalli-flores:
Replaced by https://github.com/webex/react-widgets/pull/1114 so bug can be closed.

PR #1117, opened 2020-09-04
fix(message-composer): use unique id for input

mwegman:
- Avoid multiple widgets in the DOM opening the same input element

### DIFF
--- a/packages/node_modules/@webex/react-container-message-composer/src/components/ComposerButtons.js
+++ b/packages/node_modules/@webex/react-container-message-composer/src/components/ComposerButtons.js
@@ -2,16 +2,19 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import bowser from 'bowser';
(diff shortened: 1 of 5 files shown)


packages/node_modules/@webex/react-container-message-composer/src/components/ComposerButtons.js
```diff
@@ -2,16 +2,19 @@ import React from 'react';
...
 import classNames from 'classnames';
 import bowser from 'bowser';
+import uuid from 'uuid';
 
 import {Icon} from '@momentum-ui/react';
 
 import styles from './ComposerButtons.css';
 
 function ComposerButtons({onAttachFile, composerActions}) {
+  const inputId = uuid.v4();
```
adamweeks, line 12: Could we give this a somewhat readable ID?
```suggestion
  const inputId = `messageFileInput-${uuid.v4()}`;
```

adamweeks (approved):
One question, not blocking though.

PR #1120, opened 2020-10-01
Add Drag and Drop file size limit

adamweeks:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-172895

We were checking for large files upon clicking the attach button, but drag and drop was not using the same check.

### DIFF
--- a/packages/node_modules/@webex/react-component-utils/src/files.js
+++ b/packages/node_modules/@webex/react-component-utils/src/files.js
@@ -1,14 +1,6 @@
 import uuid from 'uuid';
 
-const FILE_TYPES = {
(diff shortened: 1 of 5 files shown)


packages/node_modules/@webex/react-component-utils/src/files.js
```diff
@@ -115,3 +107,22 @@ export function getFileType(mimeType) {
 
   return 'file';
 }
+
+export function checkMaxFileSize(files, addError, removeError) {
```
lalli-flores, line 111: Do we add docs to this type of functions?

lalli-flores (approved):

PR #1126, opened 2020-10-19
fix(card-element):  enter input value for zero length of text element

mokandre:
**Description of changes made** 
* User can enter value for zero `maxLength` of the `input.text` element(https://adaptivecards.io/explorer/Input.Text.html)

**Associated issue number/tracking link**  
[SPARK-124705](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-124705)

### DIFF
--- a/packages/node_modules/@webex/react-component-md-text-input/src/index.js
+++ b/packages/node_modules/@webex/react-component-md-text-input/src/index.js
@@ -53,7 +53,7 @@ class MDTextInput extends AdaptiveCard.TextInput {
       <Input
         label={this.title}
         value={this.value}
-        maxLength={this.maxLength}
+        maxLength={this.maxLength || null}
         placeholder={this.placeholder}
         type={inputStyles[this.style].toLowerCase()}
         multiline={!!this.isMultiline}


packages/node_modules/@webex/react-component-md-text-input/src/index.js
```diff
@@ -53,7 +53,7 @@ class MDTextInput extends AdaptiveCard.TextInput {
       <Input
         label={this.title}
         value={this.value}
-        maxLength={this.maxLength}
+        maxLength={this.maxLength || null}
```
mokandre, line 56: The requirement is the user can enter input value even when `maxLength` is zero

mwegman, line 56:
```suggestion
        maxLength={this.maxLength ?? 0}
```
Are you sure you don't want this? I assume you want to default to '0'

mwegman (changes requested):

packages/node_modules/@webex/react-component-md-text-input/src/index.js
lalli-flores: @mwegman, I don't think the default value should be `0`, otherwise it will not allow for any input. By using `null` we're basically saying that we don't want to set that property at all

mokandre: @lalli-flores you are correct. @mwegman we have to allow for any input irrespective of `maxLength`

mwegman: So if `this.maxLength` is 0, you want to set it to `null` instead?

mokandre: @mwegman yes, trying to ignore `maxLength` if it is zero

mwegman (approved):

PR #1137, opened 2020-12-21
feat(w-message): display threads in activity list

mwegman:
[SPARK-192346](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-192346) #done
[SPARK-192345](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-192345) #done


packages/node_modules/@webex/react-component-activity-item/src/index.js
```diff
@@ -36,6 +36,7 @@ const propTypes = {
   }).isRequired,
   timestamp: PropTypes.string.isRequired,
   isAdditional: PropTypes.bool,
+  isReply: PropTypes.bool,
```
adamweeks, line 39: Could you add a snapshot test for `isReply = true`?

packages/node_modules/@webex/react-component-activity-list/src/index.js
```diff
@@ -102,6 +102,7 @@ function ActivityList(props) {
             isFlagged={isFlagged}
             isFlagPending={isFlagPending}
             isPending={isPending}
+            isReply={activity.parent?.type === 'reply'}
```
adamweeks, line 105: Could you make `'reply'` a constant?

packages/node_modules/@webex/react-container-activity-list/src/selectors.js
```diff
@@ -6,12 +6,37 @@ import {
...
+    getConversationNonThreadActivities
+  ],
+  (
+    threadActivities = {},
+    sortNonThreadActivities = []
+  ) => {
+    const finalList = [];
+    let start = 0;
+
+    for (let i = 0; i < sortNonThreadActivities.length; i += 1) {
```
adamweeks, line 28: This is some awesome magic! Can you throw a quick comment in about what this is doing?

packages/node_modules/@webex/react-container-read-receipts/src/selectors.js
```diff
@@ -1,7 +1,33 @@
...
+    getConversationNonThreadActivities
+  ],
+  (
+    threadActivities = {},
+    sortNonThreadActivities = []
+  ) => {
+    const finalList = [];
+    let start = 0;
+
+    for (let i = 0; i < sortNonThreadActivities.length; i += 1) {
```
adamweeks, line 19: Is it worthwhile to put this into a shared function somewhere instead of duplicating it?

packages/node_modules/@webex/widget-message/src/selector.js
```diff
@@ -1,10 +1,37 @@
...
+import {OrderedMap} from 'immutable';
 
 const getSparkInstance = (state, props) => props.sparkInstance || state.spark.get('spark');
 const getSparkState = (state) => state.spark.get('status');
 const getConversation = (state) => state.conversation;
 const getParticipants = (state) => state.conversation.get('participants');
-const getActivities = (state) => state.conversation.get('activities');
+const getConversationThreadActivities = (state) => state.conversation.get('threadActivities');
+const getConversationNonThreadActivities = (state) => state.conversation.get('sortNonThreadActivities');
+const getActivities = createSelector(
```
adamweeks, line 10: Could we store these in the convo redux store this way via the reducer so we don't have to do the computation in every selector?

adamweeks:
Since everything is in one commit, could you remove the "area" field in the commit message? `feat: display threaded messaged`

packages/node_modules/@webex/react-container-activity-list/src/selectors.js
lalli-flores: I had an old comment similar to this:

> Can you add a comment stating assumptions of these lists? e.g. are they supposed to be sorted (and by what vector)? How is the threaded activities connected to non-thread activities, etc..
This is assuming for someone that may have not spent as much time as you did understanding the convo service.

adamweeks (approved):
LGTM!

lalli-flores:
Is there a limit on how many messages get loaded? The widget keeps getting stuck at some point and it never loads more messages


stopped loading
stuck

lalli-flores:
There is also this issue that keeps happening where sometimes recent messages get displayed, messages are skipped and then it goes to the end of the activities. Then sometimes, for the same room the right messages get displayed but then it gets stuck loading as I showed in my other message
skipped messages
stuck

mwegman:
@lalli-flores when it gets stuck and won't load any more messages, what does the developer console look like? Are there any errors?

lalli-flores:
> @lalli-flores when it gets stuck and won't load any more messages, what does the developer console look like? Are there any errors?

Nothing shows up on the console, it seems to just hang

packages/node_modules/@webex/widget-message/src/container.js
```diff
@@ -136,6 +136,23 @@ export const ownPropTypes = {
   eventNames: PropTypes.object
 };
 
+const fetchMoreActivitiesIfNecessary = ({
+  activitiesStatus, conversation, firstActivity,
+  loadPreviousMessages: lpm, sparkInstance, setScrollPosition: ssp,
+  oldestPublishedDate
+}, activityList) => {
+  if (activityList?.isScrolledToTop() && firstActivity.verb !== 'create' && !activitiesStatus.isLoadingHistoryUp) {
```
adamweeks, line 144:
```suggestion
  if (activityList?.isScrolledToTop() && firstActivity.verb !== API_ACTIVITY_VERB.CREATE && !activitiesStatus.isLoadingHistoryUp) {
```

adamweeks (approved):
Changes re-reviewed and only one comment around an available constant. Not a blocker though. LGTM when you are ready!

PR #1140, opened 2021-01-26
Parse HydraId and store cluster information

taymoork2:
Fixes [SPARK-206010](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-206010)

### DIFF
--- a/packages/node_modules/@webex/react-component-utils/src/uuid-utils.js
+++ b/packages/node_modules/@webex/react-component-utils/src/uuid-utils.js
@@ -1,4 +1,40 @@
-import {base64, deconstructHydraId} from '@webex/common';
+import {constructHydraId as _constructHydraId, deconstructHydraId} from '@webex/common';
+
+/**
+ * Constructs a hydra id for a given uuid and type
+ *
+ * @export
+ * @param {string} type one of PEOPLE, TEAM, ROOM
+ * @param {any} id
+ * @param {string} cluster
+ * @returns {string}
+ */
+export function constructHydraId(type, id, cluster) {
+  return id ? _constructHydraId(type, id, cluster) : '';
+}
+
+/**
+ * Deconstructs a hydra id and provides the uuid
+ *
+ * @export
+ * @param {String} id
+ * @returns {object} hydra
+ * @returns {String} hydra.id
+ * @returns {String} hydra.type
+ * @returns {String} hydra.cluster
+ */
+export {deconstructHydraId};
+
+export const hydraTypes = {
+  ATTACHMENT_ACTION: 'ATTACHMENT_ACTION',
+  CONTENT: 'CONTENT',
+  MEMBERSHIP: 'MEMBERSHIP',
+  MESSAGE: 'MESSAGE',
+  ORGANIZATION: 'ORGANIZATION',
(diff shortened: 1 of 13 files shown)


packages/node_modules/@webex/react-component-utils/src/uuid-utils.js
```diff
@@ -34,24 +57,25 @@ export function getConvoIdFromUrl(convoUrl) {
...
     return {
-      id: string
+      id
     };
   }
 
   // Check for base 64 encoding
   try {
-    const idParts = deconstructHydraId(string);
+    const idParts = deconstructHydraId(id, cluster);
```
mwegman, line 78: `deconstructHydraId` doesn't take a second paramter

mwegman (changes requested):

packages/node_modules/@webex/react-component-utils/src/uuid-utils.js
```diff
@@ -72,27 +95,23 @@ export function validateAndDecodeId(string) {
...
- *
- * @export
- * @param {String} id
- * @returns {object} hydra
- * @returns {String} hydra.id
- * @returns {String} hydra.type
- * @returns {String} hydra.cluster
- */
-export {deconstructHydraId};
+export function constructHydraId(
```
adamweeks, line 101: `@webex/common` has a `constructHydraId`, can we just use that instead?

test/journeys/lib/hydra.js
```diff
@@ -1,8 +1,27 @@
 import {base64} from '@webex/common';
```
adamweeks, line 1: Can we use the construct and deconstruct from the sdk here?

packages/node_modules/@webex/widget-space/src/enhancers/setup.js
```diff
@@ -79,17 +79,30 @@ export function setup(props, prevProps = {}) {
...
         }
-        if (destination.type === destinationTypes.SPACEID && !spaces.getIn(['byId', validateAndDecodeId(destination.id).id])) {
-          props.fetchSpace(sparkInstance, validateAndDecodeId(destination.id));
+        if (
+          destination.type === destinationTypes.SPACEID &&
+          !spaces.getIn(['byId', validateAndDecodeId(destination.id).id])
+        ) {
+          props.fetchSpace(
+            sparkInstance,
+            validateAndDecodeId(destination.id, destination.cluster)
```
adamweeks, line 104: This function doesn't take a second param.

taymoork2: This list of if blocks was prettier'd so i could read it
the second param was already there, but i can remove it

https://github.com/webex/react-widgets/blob/master/packages/node_modules/@webex/widget-space/src/enhancers/setup.js#L92

adamweeks: I'm saying `validateAndDecodeId` doesn't take two params.

https://github.com/webex/react-widgets/blob/1310dae30e7a86f65b4a1b59e2e8d8a0a691f401/packages/node_modules/%40webex/react-component-utils/src/uuid-utils.js#L75

mwegman (approved):

PR #1147, opened 2021-03-03
fix(package): use only momentum react dependency

adamweeks:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-209358

By removing our specific dependency on momentum core and icons, allowing momentum/react to use its own dependencies, the icons work as expected finally!


packages/node_modules/@webex/webex-widget-base/src/enhancers/withCurrentUser.js
```diff
@@ -28,7 +28,9 @@ function fetchUser(props) {
     spark.getIn(['status', 'registered']) &&
     !users.get('currentUserId')
   ) {
+    /* eslint-disable no-warning-comments */
     // TODO rename sparkInstance to sdkInstance
```
taymoork2, line 32: TODOs are meant for work that will be done, not hopes and dreams

lalli-flores (approved):

PR #1151, opened 2021-03-18
Display content from threads in files widget

adamweeks:
Files widget wasn't updated with the proper area to find convo activities after the thread feature was added.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-212342

### DIFF
--- a/packages/node_modules/@webex/redux-module-conversation/src/reducer.js
+++ b/packages/node_modules/@webex/redux-module-conversation/src/reducer.js
@@ -21,6 +21,8 @@ import {filterActivities, normalizeActivities, normalizeActivity} from './helper
 
 export const initialState = fromJS({
   activities: new OrderedMap(),
+  sortNonThreadActivities: [],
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/redux-module-conversation/src/reducer.js
```diff
@@ -21,6 +21,8 @@ import {filterActivities, normalizeActivities, normalizeActivity} from './helper
 
 export const initialState = fromJS({
   activities: new OrderedMap(),
```
mwegman, line 23: You can probably remove this.

mwegman (approved):

packages/node_modules/@webex/redux-module-conversation/src/reducer.js
adamweeks: Done

taymoork2 (approved):
LGTM
image

PR #1152, opened 2021-03-22
Clicking on a space in Recents Widget creates the correct HydraID

mwegman:


adamweeks:
Can we also store `globalId` in the `constructSpaceFromHydraRoom` helper function?

PR #1156, opened 2021-04-09
fix: upgrade react-intl

mwegman:


taymoork2 (approved):
FISI

packages/node_modules/@webex/react-component-call-data-activity/src/__snapshots__/index.test.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`CallDataActivityMessage component renders properly for having a call 1`] = `
-<FormattedMessage
+<UNDEFINED
```
lalli-flores, line 4: Is this something to be concerned about??

PR #1159, opened 2021-04-29
feat(widget-space): display meeting timer in space widget menu

abailey-dev:
Add meeting timer in space widget menu.

Fixes #SPARK-223764

------------

Notes:

Noticed that the timer only works if the call is already in progress. When `call.activeParticipantsCount` is 0, the component will not update as expected. When `call.activeParticipantsCount` is initially greater than 0, the component will update as expected. Perhaps an action isn't firing as expected or perhaps incorrect logic in a reducer.

Also `call.isActive` and `call.isConnected` does not update as expected.


lalli-flores:
Replaced by https://github.com/webex/react-widgets/pull/1161

PR #1163, opened 2021-05-14
Add ability to screenshare during a meeting

taymoork2:
thanks @lalli-flores


packages/node_modules/@webex/widget-meet/src/components/call-active/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import {WebexInMeeting, WebexMeetingControls, WebexMeetingControl, WebexDataProvider} from '@webex/components';
+import {WebexInMeeting, WebexMeetingControls, WebexMeetingControl, WebexDataProvider} from '@webex/components/dist/es/webex-components.es';
```
lalli-flores, line 5: I think this is ok for debugging, but we should let webpack decide which bundle it will import.
```suggestion
import {WebexInMeeting, WebexMeetingControls, WebexMeetingControl, WebexDataProvider} from '@webex/components';
```

lalli-flores (changes requested):
The is a bug that was identified when share is taken over or when the meeting control is used to stop sharing where the info banner doesn't go away. We should prob wait for that fix so we can just update to the right version.

packages/node_modules/@webex/widget-meet/src/components/call-active/index.js
```diff
@@ -10,9 +10,10 @@ const ActiveMeeting = ({adapter, meetingId}) => (
   <WebexDataProvider adapter={adapter}>
     <div className={classNames(styles.callContainer, 'call-container')}>
       <WebexMeetingControls className={styles.callControls} meetingID={meetingId}>
-        <WebexMeetingControl type="mute-audio" />
-        <WebexMeetingControl type="mute-video" />
-        <WebexMeetingControl type="leave-meeting" />
+        <WebexMeetingControl type="mute-audio" meetingID={meetingId} />
+        <WebexMeetingControl type="mute-video" meetingID={meetingId} />
+        <WebexMeetingControl type="share-screen" meetingID={meetingId} />
+        <WebexMeetingControl type="leave-meeting" meetingID={meetingId} />
```
taymoork2, line 16: meetingID is now a required prop, which it wasnt before
This seems like a breaking change of sorts and couldnt find anything in the changelog that said it was
Was alerted when i scrolled through the console log

mwegman (approved):

packages/node_modules/@webex/widget-meet/src/components/call-active/index.js
lalli-flores: yeah, that changed. Thanks for catching it!

lalli-flores (changes requested):
The bug where the banner is not going away is now fixed 🎉

However, I noticed that the controls for interstitial show up while the meeting is loading, is there a way to avoid displaying the controls until the user has joined the meeting? we should only display the in-meeting controls when the user is in the meeting.

https://user-images.githubusercontent.com/6610308/123712842-0048e200-d828-11eb-9f54-f80029df9c1a.mp4

package.json
```diff
@@ -101,11 +101,11 @@
     "marked": "^0.7.0",
     "moment": "^2.17.1",
     "netlify": "^1.2.0",
-    "prop-types": "^15.5.10",
+    "prop-types": "^15.7.2",
     "pvutils": "^1.0.17",
-    "react": "^16.8.4",
+    "react": "^16.14.0",
```
lalli-flores, line 106: The latest version of the components [has been updated to use React v17](https://github.com/webex/components/blob/master/package.json#L38)

lalli-flores:
@taymoork2, please be sure to update tests as necessary. Seems like the failing journey test is relevant to the code changes
image

lalli-flores (changes requested):
Versions `@webex/sdk-component-adapter": "1.35.0` & `@webex/components": "1.69.1` seem to be versions with the bug where screen share banner is not going away after stopping share. Can you confirm?

Edit: I think fix is in [SDK adapter v1.36.1](https://github.com/webex/sdk-component-adapter/blob/master/CHANGELOG.md#1361-2021-06-24). Please verify with Endava that that is the correct version

lalli-flores (approved):
From my testing, everything seems to be working 🥳

PR #1164, opened 2021-05-18
fix(rc-activity-item-base): make sure correct flag icon is used

taymoork2:
Highlight message follow-up flag

Fixes: [SPARK-209503](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-209503)


lalli-flores (approved):
I tested it and it seems to work. The color is not blue, but I think it's ok since the UI seems to have changed in the clients as well.

This is what it looks like in macOS client:
image

PR #1165, opened 2021-05-19
feat(dangerouslyIgnoredRequiredPlugins): allow users to disable check…

zlampertcisco:
… for required plugins

This change is motivated by the need to inject an sdk instance that may not have certain plugins available to it. For example, plugin-phone is being deprecated but is still required for the widgets. This approach will allow for users to "dangerously" ignore such plugins when implementing the widget without introducing potential breaking changes to legacy instances of the widget.

PR #1166, opened 2021-05-24
feat: remove phone plugin

mwegman:


lalli-flores (changes requested):
This should be a `fix` or `feat` commit, as we are changing the dependencies necessary for the widget to run.
@zlampertcisco has a similar PR https://github.com/webex/react-widgets/pull/1165 because of issues with the phone plugin, so I think they will need the version bump

lalli-flores (approved):
All tests pass 🙌🏼

PR #1168, opened 2021-05-28
update SDK for conversationsSummary resource

taymoork2:
Use `/conversationsSummary` API instead of `/conversations` as it has a looser rate-limit when loading list of rooms.


lalli-flores (changes requested):
Seems like unit and journey tests are not passing. Running again in case of flakiness but we should investigate if the SDK update broke something.

Your commit on the API call should be a `feat` since we want to trigger a release.

lalli-flores (approved):

lalli-flores:
Some tests are not passing due to Momentum buttons not showing up, but will be addressed by https://github.com/webex/react-widgets/pull/1163

PR #1172, opened 2021-06-24
fix(widget-roster): people tab scrollbar fix

chetangoel01:
Fixes an issue where part of the native scrollbar in the participant list gets hidden when there are external participants.
scroll_issue


lalli-flores (changes requested):
I don't think this is the right fix, the banner within the participants tab is taking up all the space instead of staying at the top and leaving the rest of the space for the participant list
image

lalli-flores (approved):
Looks good, thanks for confirming it also works with Edge @chetangoel01!

PR #1173, opened 2021-07-08
Add ability to turn off presence

suphs-cisco:
Adding a prop so that users may disable the presence feature of the Space widget. Disabling presence could optimize widget load time, as there will be less APIs to call.


lalli-flores:
@suphs-cisco, thank you for taking the time to submit optimizations to the code!
Did you get a chance to [review our contribution guide](https://github.com/webex/react-widgets/blob/master/CONTRIBUTING.md#contributing-code)? Seems like your commits are not aligning with our guidelines.
Additionally, if you could add a description in the pull request about the motivation for this changes, it would be appreciate it, as it provides a record.

lalli-flores (changes requested):
I think everything looks good. I did some testing, and seems like things are working as expected. 
Minor comments + the commit fixes and we should be good to go

lalli-flores (approved):

PR #1182, opened 2021-08-16
fix(style): remove extra padding between avatar and text

chetangoel01:
Padding between Avatar and Copy should be 16px. There is 16px spacing but individual components like avatar and copy both have some padding that adds to this and it goes beyond 20px. Removed left padding on texts to make padding 16px again.
[SPARK-219839](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-219839)


chetangoel01:
previously: before


now: after

taymoork2 (approved):
LGTM

PR #1183, opened 2021-08-16
fix(userAgent): Added changes to identify user agent for widgets.

jonniespratley:


jonniespratley:
Update snapshots, updated specs to pass in name of component.

mwegman (approved):

PR #1190, opened 2021-08-24
fix(widget-space): fixed startCall issue

emanuallan:
I added a check for the sdkAdapter before placeCall

### DIFF
--- a/test/journeys/specs/space/startup-settings.js
+++ b/test/journeys/specs/space/startup-settings.js
@@ -37,6 +37,27 @@ describe('Space Widget Startup Settings Tests', () => {
     });
   });
 
+  describe('startCall', () => {
+    it.skip('call is placed after startCall is set to true', () => {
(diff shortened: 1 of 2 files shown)


mwegman (approved):

test/journeys/specs/space/startup-settings.js
```diff
@@ -37,6 +37,27 @@ describe('Space Widget Startup Settings Tests', () => {
     });
   });
 
+  describe('startCall', () => {
+    it('call is placed after startCall is set to true', () => {
```
taymoork2, line 41: Keep the test, but skip it, because we can't verify if it works or not due to issues with services at the time of my review
```suggestion
    it.skip('call is placed after startCall is set to true', () => {
```

taymoork2 (approved):
LGTM

mwegman (approved):

PR #1193, opened 2021-09-02
doc(README): update package serve command

jonniespratley:


lalli-flores (approved):
Thanks for updating the docs

PR #1197, opened 2021-09-13
Fix(SPARK-219844): Message Spacing

emanuallan:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-219844


emanuallan:
Will not be continued for the time being.

PR #1198, opened 2021-09-16
 feat(space-widget): add optional props showSubmitButton & sendMessageOnReturnKey

zaklampert2:
Allow users to use a send button and disable the return key for sending message. Show an error if configured in a way that will not allow users to send a message.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-267233


lalli-flores (changes requested):
Something is not right with the defaults, as the widget is not loading at all on the demo page.
image

Also, please use the widget name as the scope of the commit, not the Jira number. Jira wouldn't mean anything on [the changelog](https://github.com/webex/react-widgets/blob/master/CHANGELOG.md)

```
feat(space-widget): add optional props showSubmitButton & sendMessageOnReturnKey

zlampertcisco:
@lalli-flores updated to pass new props through the setup enhancer's selector and also changed the commit's scope.

PR #1201, opened 2021-09-27
fix(sdk-adapter): upgrade to 1.63.0

jonniespratley:
Updated adapter version to latest and verified that `widget-spaces` and `widget-recents` widgets are working correctly.

- Tested demo page widgets for recent with all options checked.
- Tested demo page spaces with options checked.


packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -207,13 +207,15 @@ const getMeeting = ({sdkAdapter: {meetingsAdapter}, destination}) => new Promise
...
+
         // eslint-disable-next-line no-param-reassign
         meetingsAdapter.meetings[meeting.id] = {
           ID: meeting.id,
+          ...m,
           title,
           localAudio,
           localVideo,
-          localShare: null,
           remoteAudio: null,
```
lalli-flores, line 219: Since you are now taking the meeting object from the adapter, remove this other values and just overwrite `title`, `localAudio` and `localVideo`, as those are the only things that need to be overwritten.

jonniespratley:
Going to issue a new PR #1204

PR #1202, opened 2021-09-28
Fix meeting related onEvent events

taymoork2:
A multitude of fixes and updates to the meeting + meet component implementation in the widget to reenable `calls:created` and `calls:connected` events being emitted by the widget.
Also major refactor to use the correct properties of the locus object.
Removed some code that no longer works/exists in the meeting plugin that was being used still in the widgets


packages/node_modules/@webex/redux-module-media/src/actions.js
```diff
@@ -314,21 +314,19 @@ function handleCreatedCalls(sparkInstance, dispatch) {
   };
   const getMeetingById = ({id}) => sparkInstance.meetings.getMeetingByType('id', id);
 
-  sparkInstance.meetings.on('meeting:added', ({meeting, type}) => {
-    if (['JOIN', 'INCOMING'].find((TYPE) => type === TYPE)) {
-      meeting.members.on('members:update', () => {
```
taymoork2, line 319: @arun3528 can you explain why this was causing issues

lalli-flores (approved):

PR #1204, opened 2021-10-04
fix(sdk-adapter): upgrade to ^1.64.0

jonniespratley:
Updated adapter version to latest and verified that `widget-spaces` and `widget-recents` widgets are working correctly.

- Tested demo page widgets for recent with all options checked.
- Tested demo page spaces with options checked.


lalli-flores (changes requested):
This doesn't look good at all. Did you add the components CSS?
You also need to add the class `wcx-theme-light` to the widget meet (or whichever is the parent div for where the components render)

jonniespratley:
> This doesn't look good at all. Did you add the components CSS? You also need to add the class `wcx-theme-light` to the widget meet (or whichever is the parent div for where the components render)

Should the consumer of the component set the base theme class?

jonniespratley:
> > This doesn't look good at all. Did you add the components CSS? You also need to add the class `wcx-theme-light` to the widget meet (or whichever is the parent div for where the components render)
> 
> Should the consumer of the component set the base theme class?

Also for the css variables theming, there should be fallback colors incase browsers that do not support CSS vars.

jonniespratley:
This PR updates adapter and fixes styling issues when upgrading.

jonniespratley:
> You missed this [update on the meeting object setup](https://github.com/webex/react-widgets/pull/1204/files#diff-336290fc6b9f66707ab84ac864149c4c3dc662a592c3bf933a662b6164064135L102)

Fixed

PR #1220, opened 2022-02-24
Correctly grab conversationUrl from meetings payload

taymoork2:
[SPARK-275103](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-275103)


thescripted (approved):
Looks good, ran through a few test cases and checked that the space always re-opens when the caller dials again:
- Caller cancels the call
- Callee rejects the call
- Caller hangs up after they're connected
- Callee hangs up after they're connected

PR #1284, opened 2023-03-28
fix(crossLaunchToken): crossLaunchToken

sokn-sys:
Sending the cross launch token to the croslaunch service


akulakum (approved):

nvanganm (approved):
look good.

PR #1327, opened 2023-07-14
fix(sdTruncationFix): speeddial truncation fix

akulakum:
Description:  Fixed the speed dial truncation issue and removed title from the Avatar tag because already there is a title at parent level in the speed dial item component


packages/node_modules/@webex/widget-speed-dial/src/SpeedDialItem.tsx
```diff
@@ -167,7 +167,7 @@ export const SpeedDialItem = ({
                 {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
                 <div className={sc('actions')}>
                   <Flex className={sc('avatar')}>
-                    <Avatar initials={removeBracketsAndContent(removeCommaIfNeeded(item?.displayName))} size={48} title={item?.displayName} src={item?.photo} />
```
Waseemraj, line 170: No need of tooltip?

Waseemraj (approved):

PR #1337, opened 2023-10-20
feat(): space id code change in space widget and demo

Neeraj-swarnkar:
# JIRA - [SPARK-457886](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-457886)

# Changes -
Description:

Modified the space ID code within the space widget to throw an error and tooltip 
Default values for enableUnifiedMeetings and enableAdhocMeetings have been set to true. Configuration changes are now migrated to V2 meetings. 
Upgraded the webex dependencies
Additionally, updated the demo samples.

Screenshots:
Screenshot 2023-10-19 at 2 30 55 PM

Screenshot 2023-10-25 at 4 13 44 PM

image

PR #1388, opened 2024-06-10
fix(space-widgets): fix for joining meeting using SIP destination

Kesari3008:
After the change done as part of PR: https://github.com/webex/react-widgets/pull/1337, joining meetings using SIP URI stopped working.
Jira:  https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505404

This happened due to the check added for fetching the meeting info based on the conversation URL which was impacting the SIP URI cases as well for guest users. 
Fixed the check and it is working for SIP URI but not for the conversation URL

### DIFF
--- a/packages/node_modules/@webex/widget-space/src/container.js
+++ b/packages/node_modules/@webex/widget-space/src/container.js
@@ -11,8 +11,6 @@ import LoadingScreen from '@webex/react-component-loading-screen';
 import Timer from '@webex/react-component-timer';
 import ErrorDisplay from '@webex/react-component-error-display';
 
-import {CONSTANTS} from '@webex/plugin-meetings';
-
 import {unregisterDevice} from '../../react-redux-spark/src/actions';
 
 import ActivityMenu from './components/activity-menu';
@@ -112,7 +110,6 @@ export class SpaceWidget extends Component {
     } = props;
     const {formatMessage} = props.intl;
     let errorElement;
-    const currentConversation = conversation.getIn(['url']);
 
     if (errors.get('hasError') || conversation.getIn(['status', 'error'])) {
       let widgetError = errors.get('errors').first();
@@ -148,10 +145,7 @@ export class SpaceWidget extends Component {
       const {preferredWebexSite} = sparkInstance.meetings;
       let isMeetButtonDisabled = false;
 
-      if (sparkInstance.meetings.getMeetingByType(CONSTANTS.CONVERSATION_URL, currentConversation)) {
-        isMeetButtonDisabled = false;
-      }
-      else {
+      if (props.destinationType === 'spaceId' && currentActivity === 'meet' && activityTypes[3].name === 'meet') {
         isMeetButtonDisabled = !preferredWebexSite;
       }
 
@@ -236,7 +230,7 @@ export class SpaceWidget extends Component {
             {menuButton}
           </div>
           <div className={classNames('webex-widget-body', styles.widgetBody)}>
-            {isMeetButtonDisabled && currentActivity === 'meet' ? errorElementSpace : widgets}
+            {isMeetButtonDisabled ? errorElementSpace : widgets}
           </div>
         </div>
       );


sreenara (changes requested):
Can we expand on all the tests attempted? 
Also, I think we should be able to create a meeting using conversation URL. Please check the kitchen sink page where this is a way to create a meeting in a space.
I think we are only supposed to block based on roomId. See this [developer API](https://developer.webex.com/docs/api/v1/rooms/list-rooms) for rooms which is different from conversationURL.

Kesari3008:
> Can we expand on all the tests attempted? Also, I think we should be able to create a meeting using a conversation URL. Please check the kitchen sink page where this is a way to create a meeting in a space. I think we are only supposed to block based on roomId. See this [developer API](https://developer.webex.com/docs/api/v1/rooms/list-rooms) for rooms which is different from conversationURL.

In space-widgets, option to join using the conversation URL is not present. We are only blocking roomID scenario.

packages/node_modules/@webex/widget-space/src/container.js
```diff
@@ -148,10 +145,7 @@ export class SpaceWidget extends Component {
       const {preferredWebexSite} = sparkInstance.meetings;
       let isMeetButtonDisabled = false;
 
-      if (sparkInstance.meetings.getMeetingByType(CONSTANTS.CONVERSATION_URL, currentConversation)) {
-        isMeetButtonDisabled = false;
-      }
-      else {
+      if (props.destinationType === 'spaceId' && activityTypes.some((activity) => activity.name === 'meet')) {
```
Neeraj-swarnkar, line 148: How is the flow? if user trying to join with CONVERSATION_URL ??
Earlier we were hiding the meet button based on the conditions, those configuration changes were for migrating to V2 meetings.

Kesari3008: In space-widgets, I don't see the option of joining with CONVERSATION_URL. Could you please help me with steps for the same?
The existing condition was blocking flow for SIP URI but not for SpaceID so I updated the condition here to reverse that.

Neeraj-swarnkar:
image
Can you verify if within the space widget an error and tooltip both shows correct ?

packages/node_modules/@webex/widget-space/src/container.js
Neeraj-swarnkar: It's the same as the way we have in the KS samples sdk app, in KS app also, join button disabled, 

image

Kesari3008:
> image Can you verify if within the space widget an error and tooltip both shows correct ?

Updated the screenshot where tooltip is also visible

Neeraj-swarnkar:
> > image Can you verify if within the space widget an error and tooltip both shows correct ?
> 
> Updated the screenshot in the description where tooltip is also visible

Thanks @Kesari3008

packages/node_modules/@webex/widget-space/src/container.js
Kesari3008: In the kitchen sink app, the dropdown has a conversation URL as one of the options, there is nothing similar in space widgets. I only see 4 options to be chosen as destination type: To Space, To Person, To PersonId, To SIP URI. Now which one of these would be the case for conversation URL?

Screenshot 2024-06-11 at 5 13 34 PM

Neeraj-swarnkar: oh yes, we don't have convURL here, You can choose any options and try passing the convURL and check,? In SDK there's utility meeting-info which figure out based on input what type of destination it is. Just to test once..

Kesari3008: I have tried choosing the 'To Space' option and entered the conversation URL and it says unable to load space.

sreenara (approved):
Let's keep the investigation to add testing for a future task.

Neeraj-swarnkar (approved):
Looks Good.

PR #1397, opened 2024-06-21
fix(spark:526772): callhistory date annuouncement

sokn-sys:
Description:
This PR contains voiceover announcements for the call history date and in the voicemail. If there is no data, a comma should not be added
Vidcast link
https://app.vidcast.io/share/ea4598fd-1090-42d8-800f-ea0b1f62c053

ch1

ch2

### DIFF
--- a/packages/node_modules/@webex/widget-call-history/src/CallHistoryItem.tsx
+++ b/packages/node_modules/@webex/widget-call-history/src/CallHistoryItem.tsx
@@ -18,6 +18,7 @@ import {
 import {
   formatDate,
   formatDateDDMMYYYY,
+  formatDateForAnnouncement,
   formatDurationFromSeconds,
   formatTime,
   formatTimeToSupport24Hours
@@ -112,7 +113,7 @@ export const CallHistoryItem = ({
       className={cssClasses}
       onPress={onPress}
       data-testid="call-history-item"
-      aria-label={`${(name)}, ${(phoneNumber && formatPhoneNumberForAnnouncement(phoneNumber))}, ${(callingSpecific ? callingSpecific + ',' : '')} ${isOutgoing ? t('outGoingCallText') : isMissed ? t('missedCallText') : t('incomingCallText')}, ${(formatDurationForAnnouncement(durationSeconds as number))}, ${(startTime && isLocaleGerman ? formatDateDDMMYYYY(startTime) : formatDate(startTime))}, ${(startTime && isLocaleGerman ? formatTimeToSupport24Hours(startTime) : formatTime(startTime))}, ${t('recentCallsFocusButton')}`}
+      aria-label={`${(name)}, ${(phoneNumber && formatPhoneNumberForAnnouncement(phoneNumber))}, ${(callingSpecific ? callingSpecific + ',' : '')} ${isOutgoing ? t('outGoingCallText') : isMissed ? t('missedCallText') : t('incomingCallText')}, ${(formatDurationForAnnouncement(durationSeconds as number))}, ${(formatDateForAnnouncement(startTime))}, ${(startTime && isLocaleGerman ? formatTimeToSupport24Hours(startTime) : formatTime(startTime))}, ${t('recentCallsFocusButton')}`}
     >
       <ListItemBaseSection position="start" className={sc('icon')}>
         <Avatar initials={chInitials} title={chTitle} size={32} />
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/widget-call-history/src/utils/dateUtils.ts
```diff
@@ -127,3 +127,37 @@ export function formatDateDDMMYYYY(date: string) {
...
+ * @returns {string} A announced date as in "Today", "Yesterday", or "june 10,2024"
+ */
+export const formatDateForAnnouncement = (dateValue: string) => {
+  const date = new Date(dateValue);
+  const today = new Date();
+  const yesterday = new Date();
+  yesterday.setDate(yesterday.getDate() - 1);
+
+  if (isSameDate(date, today)) {
+      return "Today";
```
Waseemraj, line 143: How do we manage for other languages?

packages/node_modules/@webex/widget-call-history/src/utils/dateUtils.ts
```diff
@@ -127,3 +127,37 @@ export function formatDateDDMMYYYY(date: string) {
...
+  }
+  if (isSameDate(date, yesterday)) {
+      return "Yesterday";
```
Waseemraj, line 146: Same question here also.
How do we manage for other languages?

packages/node_modules/@webex/widget-call-history/src/CallHistoryItem.tsx
```diff
@@ -112,7 +113,7 @@ export const CallHistoryItem = ({
       className={cssClasses}
       onPress={onPress}
       data-testid="call-history-item"
-      aria-label={`${(name)}, ${(phoneNumber && formatPhoneNumberForAnnouncement(phoneNumber))}, ${(callingSpecific ? callingSpecific + ',' : '')} ${isOutgoing ? t('outGoingCallText') : isMissed ? t('missedCallText') : t('incomingCallText')}, ${(formatDurationForAnnouncement(durationSeconds as number))}, ${(startTime && isLocaleGerman ? formatDateDDMMYYYY(startTime) : formatDate(startTime))}, ${(startTime && isLocaleGerman ? formatTimeToSupport24Hours(startTime) : formatTime(startTime))}, ${t('recentCallsFocusButton')}`}
```
Waseemraj, line 115: Doesn't it break for German users?

packages/node_modules/@webex/widget-call-history/src/utils/dateUtils.ts
```diff
@@ -127,3 +127,37 @@ export function formatDateDDMMYYYY(date: string) {
     year: 'numeric',
   });
 }
+
+/**
+ * @description Handle formatting the date for announce.
+ * @param {string} date The date string to format
+ * @returns {string} A announced date as in "Today", "Yesterday", or "june 10,2024"
+ */
+export const formatDateForAnnouncement = (dateValue: string) => {
```
akulakum, line 136: I've reviewed the formatDateForAnnouncement method in the PR and noticed that it currently returns "Today" and "Yesterday" as static strings. This approach does not support localization, which could be an issue for users in different locales.


Fortunately, we have an existing formatDate method that handles localization for "today" and "yesterday" by using d.toRelativeCalendar({}) from the Luxon library, which is designed to support localization.


To ensure consistency and to take advantage of Luxon's localization support, I suggest we update the formatDateForAnnouncement method as follows:

export const formatDateForAnnouncement = (date: string) => {
  const d = getDateTimeFromString(date);

  if (
    d.toISODate() === DateTime.now().toISODate() ||
    d.toISODate() === DateTime.now().minus({ days: 1 }).toISODate()
  ) {
    return d.toRelativeCalendar({});
  }

  return d.toLocaleString({
    month: 'long', // Use 'long' for full month name
    day: 'numeric',
    year: 'numeric', 
  });
}
This updated method will now return localized strings for "today" and "yesterday" and will format other dates in a more accessible manner for announcements.

packages/node_modules/@webex/widget-call-history/src/utils/dateUtils.ts
sokn-sys: will be updating the code to supports other language

packages/node_modules/@webex/widget-call-history/src/CallHistoryItem.tsx
sokn-sys: We have already taken care of displaying the date for German users. This announcement is in accordance with IVY comments; 
we need to announce the date, for example, June 20th, 2024

packages/node_modules/@webex/widget-call-history/src/utils/dateUtils.ts
sokn-sys: Sure will check for other language also

Waseemraj (approved):

akulakum (approved):

PR #1398, opened 2024-06-21
fix(speeddial): phone number regex change

sayachat-webex:
This PR is to change regex, used for validation check of phone number in speed dial add and edit.
1) Will not limit the length of numbers in the speed dial input field. There will be a minimum length of 1.
2) Allow special chars to be saved in between the numbers. (this is for example used when dialing into Webex meetings or other IVR-based systems. That interacts with DTMF tones. e.g. "," would insert 2sec pause")

[https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-535963](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-535963)

### DIFF
--- a/packages/node_modules/@webex/widget-speed-dial/src/SpeedDialForm.tsx
+++ b/packages/node_modules/@webex/widget-speed-dial/src/SpeedDialForm.tsx
@@ -354,7 +354,7 @@ export const SpeedDialForm = ({
                 rules={{
                   required: t('form.phone.required'),
                   pattern: {
-                    value: /^[+]?[(]?\d{3}[)]?[-\s.]?\d{3}[-\s.]?\d{4,9}$/,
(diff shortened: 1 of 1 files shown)


Waseemraj (approved):

sayachat-webex:
> I noticed that we're currently not limiting the length of numbers in the speed dial input field. Could you please confirm if we are still displaying a validation error for exceeding the maximum length of numbers?
> 
> Additionally, if the phone number field is not a required field, I suggest we consider removing the asterisk (*) symbol in front of the "Phone" label to avoid confusion.

The validation error we are showing is  "Must be a valid number". We are not validating the length, but we are validating if the string contains digits and special characters and not alphabets. 

We can remove the asterisk (*)

Waseemraj (approved):

PR #1415, opened 2024-08-01
fix(spacewidget): add focus to the attach file button

Kesari3008:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-539198 

Problem: The Choose a file button in the chat panel doesn't visually appear, focusable using the Tab key, and selectable using the Enter key to choose a file so the user cannot identify the functionality of the control.

Fix: Added the Button element to bring the keyboard focus for the input element used to attach files.

Screenshot 2024-08-02 at 9 59 29 PM


adhmenon:
Priya, could you attach a ss or a gif of the changes? Just to understand the change better.

Kesari3008:
> Priya, could you attach an ss or a gif of the changes? Just to understand the change better.

@adhmenon  Screenshot is uploaded

adhmenon (approved):
LGTM.

PR #1421, opened 2024-08-08
fix(ActivityShare): accessibility for download buttons

mkesavan13:
### This pull request addresses
This PR addresses the issue of download buttons not being accessible. The download buttons can be accessed through `Shift + Tab` but the image files' Download button was not accessible.

### by making the following changes
- To fix it, we have disabled the `hover` ability of the Image component rather show the file name and downlaod button by default. This makes the download button present to be accessible by doing `Shift + Tab`
- Additionally, I've added the file names along with the download button's aria-label so that it doesn't simply read Download but rather what to download.

### Screen recording

https://github.com/user-attachments/assets/bcbca58d-c9a1-4d6b-8142-36254226edba

### DIFF
--- a/packages/node_modules/@webex/react-component-activity-share-file/src/__snapshots__/index.test.js.snap
+++ b/packages/node_modules/@webex/react-component-activity-share-file/src/__snapshots__/index.test.js.snap
@@ -57,7 +57,7 @@ exports[`ActivityShareFile component renders properly 1`] = `
       className="webex-share-action-item shareActionItem"
     >
       <ForwardRef(render)
-        ariaLabel="Download"
+        ariaLabel="Download testImage.js"
         circle={true}
         onClick={[Function]}
         size={32}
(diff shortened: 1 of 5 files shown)


Kesari3008 (approved):
Why does it say dimmed while reading the aria-label and what does it mean?
Did we check if updating the css for md_content is not impacting any other element?

adhmenon (approved):
Overall LGTM. That 'dimmed' issue while reading the aria label is mostly due to the fact that the button is disabled.

mkesavan13:
@adhmenon @Kesari3008 - Thanks for the approvals.

I think the `dimmed` readout is because this is a light-themed Widget. The button isn't disabled. I'll go ahead and merge the PR once the checks pass

mkesavan13:
> Why does it say dimmed while reading the aria-label and what does it mean? Did we check if updating the css for md_content is not impacting any other element?

@Kesari3008 - For the second one, no. Since we have this CSS encapsulated under a parent CSS style that is component-specific, we do not have any other component that is affecting because of this change.

PR #1424, opened 2024-08-13
React Widgets: Fixed message action buttons not showing on hover and key order

adhmenon:
# Problem
* Currently, in the react widgets when we have focus on the messages it does not show the action items and only shows them on hover.
* This is an issue for users on mobile and desktop as they cannot access these items nor do they have instructions to observe them.
* When we click on 'delete', the dialog window was also not keyboard focusable.

# Fix
* Code has been added so that the buttons are now visible and focusable when the message gets focus.
* The navigation is as expected in desktop and should be in mobile browsers as well (currently testing this). This clears the Applause requirements.
* The dialog window is now keyboard focusable.

## JIRA
[SPARK-549503](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-549503)
[SPARK-550134](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-550134)

### Showing the buttons visible and focusable
button-visible

### Showing dialog window focusable
dialog-alert

### Accessibility Label for 'Delete'
Screenshot 2024-08-14 at 11 12 51 AM

### Accessibility Label for 'Flag'
Screenshot 2024-08-14 at 11 12 40 AM

### Accessibility Label for 'Selected Flag'
Screenshot 2024-08-14 at 3 58 23 PM

### DIFF
--- a/packages/node_modules/@webex/react-component-activity-item-base/src/index.js
+++ b/packages/node_modules/@webex/react-component-activity-item-base/src/index.js
@@ -95,6 +95,7 @@ function ActivityItemBase({
           color="black-100"
           name={isFlagged ? 'flag-active_16' : 'flag_16'}
           onClick={handleOnFlag}
+          aria-pressed={isFlagged}
         />
       </div>
     );
--- a/packages/node_modules/@webex/react-component-activity-item-base/src/styles.css
+++ b/packages/node_modules/@webex/react-component-activity-item-base/src/styles.css
@@ -130,7 +130,3 @@
 .activityItem:focus-within .activityPostActions {
   visibility: visible;
 }
-
-.visually-hidden {
-  visibility: hidden;
-}
--- a/packages/node_modules/@webex/react-component-activity-item-base/src/styles.css
+++ b/packages/node_modules/@webex/react-component-activity-item-base/src/styles.css
@@ -126,6 +126,11 @@
   visibility: visible;
 }
 
-.activityItem:hover .activityPostActions {
+.activityItem:hover .activityPostActions,
+.activityItem:focus-within .activityPostActions {
   visibility: visible;
+}
+
+.visually-hidden {
+  visibility: hidden;
 }
--- a/packages/node_modules/@webex/react-component-confirmation-modal/src/__snapshots__/index.test.js.snap
+++ b/packages/node_modules/@webex/react-component-confirmation-modal/src/__snapshots__/index.test.js.snap
@@ -3,7 +3,10 @@
 exports[`ConfirmationModal component renders properly 1`] = `
 <div>
   <div
+    aria-modal="true"
     className="webex-dialogue-modal dialogueModal"
+    role="dialog"
+    tabIndex="-1"
   >
     <div
       className="webex-dialogue-modal-body dialogModalBody"
--- a/packages/node_modules/@webex/react-component-confirmation-modal/src/index.js
+++ b/packages/node_modules/@webex/react-component-confirmation-modal/src/index.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect, useRef} from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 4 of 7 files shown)


packages/node_modules/@webex/react-component-confirmation-modal/src/__snapshots__/index.test.js.snap
```diff
@@ -3,7 +3,10 @@
 exports[`ConfirmationModal component renders properly 1`] = `
 <div>
   <div
+    aria-modal="true"
```
adhmenon, line 6: Had to make changes to the snapshot in order for tests to pass.

packages/node_modules/@webex/react-component-confirmation-modal/src/index.js
```diff
@@ -46,7 +54,13 @@ function ConfirmationModal({
 
   return (
     <div>
-      <div className={modalClassNames}>
+      <div
+        ref={modalRef}
```
adhmenon, line 58: Here I have added an `aria-modal` so it is understood that the window is a modal window. The user then knows how to interact with it.

packages/node_modules/@webex/react-component-confirmation-modal/src/index.js
```diff
@@ -30,6 +30,14 @@ function ConfirmationModal({
 }) {
   const modalClassNames = classNames('webex-dialogue-modal', styles.dialogueModal);
 
+  const modalRef = useRef(null);
+
+  useEffect(() => {
+    if (modalRef.current) {
+      modalRef.current.focus();
```
mkesavan13, line 37: We could rather use `<FocusLock/>` and `autoFocus` attrib there. That will let us remove all the `refs`

packages/node_modules/@webex/react-component-activity-item-base/src/styles.css
```diff
@@ -126,6 +126,7 @@
   visibility: visible;
 }
 
-.activityItem:hover .activityPostActions {
+.activityItem:hover .activityPostActions,
+.activityItem:focus-within .activityPostActions {
```
mkesavan13, line 130: Smart fix 😎

mkesavan13 (changes requested):
Just one comment.

packages/node_modules/@webex/react-component-confirmation-modal/src/__snapshots__/index.test.js.snap
Shreyas281299: Did you manually fixed it or used the command?

packages/node_modules/@webex/react-component-confirmation-modal/src/index.js
Shreyas281299: @mkesavan13 since space-widgets is using node14 we can't use focus-lock right now. This is a way to solve this but I agree a better fix is to use focus-lock, but it is dependent on node version

Shreyas281299 (approved):

packages/node_modules/@webex/react-component-confirmation-modal/src/__snapshots__/index.test.js.snap
adhmenon: Had to manually add it... tried the -u command but it didn't update the code.

mkesavan13 (approved):
Looks good to me. Approved

packages/node_modules/@webex/react-component-activity-item-base/src/index.js
```diff
@@ -95,6 +95,7 @@ function ActivityItemBase({
           color="black-100"
           name={isFlagged ? 'flag-active_16' : 'flag_16'}
           onClick={handleOnFlag}
+          aria-pressed={isFlagged}
```
adhmenon, line 98: This is so that when the flag is active it announces 'selected', as per Applause requirements.

PR #1426, opened 2024-08-19
fix(spacewidget): fix reflow of content in 320px wide browsers

bhabalan:
### Problem
Currently, in the react space widget, when the browser width is at 320px, alot of items get cropped on screen.
This is an issue for users on smaller screens as they cannot access all the content and controls.

### Fix
CSS fixes have been added to the activity component and its children for 320px so that all the content and controls are visible and available to the user

JIRA: [SPARK-549499](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-549499)

Vidcast: https://app.vidcast.io/share/7c9cb830-9d15-44e9-87c5-384c5ed224d3

Before:
image

After:
image


packages/node_modules/@webex/react-component-activity-share-files/src/index.js
```diff
@@ -9,6 +9,7 @@ import AdaptiveCard from '@webex/react-component-adaptive-card';
 import {CARD_CONTAINS_IMAGE} from '@webex/react-component-utils';
 
 import styles from './styles.css';
+import './momentum.scss';
```
mkesavan13, line 12: So, until we import here, the momentum overrides won't take effect. But after this, the CSS is going to stay in the `CSSOM` right?

Won't that override the CSS affecting the application where the Widget would reside? (Assuming that app also uses momentum library)

packages/node_modules/@webex/react-component-activity-share-files/src/styles.css
```diff
@@ -4,7 +4,7 @@
 }
 
 .shareList {
-  width: 320px;
+  max-width: 320px;
```
mkesavan13, line 7: Won't it feel inconsistent if the width of shared file A and shared file B is different based on the length of their names?

packages/node_modules/@webex/react-component-activity-text/src/styles.css
```diff
@@ -11,6 +12,14 @@
   transition: color 0.3s, margin-left 0.3s ease 0.5s;
 }
 
+@media (max-width: 320px) {
+  /* Reply text shouldn't get cropped */
```
mkesavan13, line 16: In which case shouldn't we having a word-wrap here?

mkesavan13 (changes requested):
Mostly looks good. Only a few comments.

.gitattributes
```diff
@@ -1,2 +1,3 @@
 /node_modules linguist-vendored
 /packages/node_modules linguist-vendored=false
+/packages/node_modules/** linguist-generated=false
```
mkesavan13, line 3: Thanks for that 😎

packages/node_modules/@webex/react-component-activity-item-base/src/styles.css
```diff
@@ -100,6 +110,15 @@
   margin-left: 56px;
 }
 
+@media (max-width: 320px) {
+.activityItem.additional {
+  /*
+   * reduce wasted space for smaller screens
+   */
+  margin-left: 48px;
+}
```
mkesavan13, line 119: nitpick: Could you please indent this?

packages/node_modules/@webex/react-component-activity-share-files/src/styles.css
```diff
@@ -7,6 +7,65 @@
...
+    max-width: 100%;
+    width: unset;
+  }
+
+  .shareList div[class*="md-content__chat-inner-container"] {
+    width: unset;
+    min-width: 170px;
+  }
+
+  .shareList div[class=" md-content-file--full"] {
```
mkesavan13, line 21: Why do we have a space here?
```suggestion
  .shareList div[class="md-content-file--full"] {
```

mkesavan13 (approved):
A couple of nitpick comments. Looks good otherwise.

**Approving**

packages/node_modules/@webex/react-component-activity-share-files/src/styles.css
bhabalan: It's an exact match only if there is a preceding space
image

Shreyas281299 (approved):

PR #1431, opened 2024-09-02
fix(space-widget): links in messages should open in new tab

bhabalan:
**Problem:**
When the space widget is loaded inside of an iframe, the links in the messages when clicked are restricted from being loaded.

**Solution:**
Add target="_blank" attribute to open these links in a new tab

JIRA: SPARK-556112

### DIFF
--- a/packages/node_modules/@webex/react-component-activity-text/src/index.test.js
+++ b/packages/node_modules/@webex/react-component-activity-text/src/index.test.js
@@ -6,6 +6,7 @@ import ActivityText from '.';
 describe('ActivityText component', () => {
   const displayName = 'Test Text';
   const cleanContent = '<em>Test Text</em>';
+  const cleanLink = '<a href="https://example.com">Test Text</a>';
   const dirtyContent = '<script>alert(\'dirtyContent\')</script>';
 
   it('renders clean text properly', () => {
@@ -16,6 +17,12 @@ describe('ActivityText component', () => {
 
   it('renders clean dangerous text properly', () => {
     const component = renderer.create(<ActivityText content={cleanContent} />);
+
+    expect(component).toMatchSnapshot();
+  });
+
+  it('renders clean links properly', () => {
+    const component = renderer.create(<ActivityText content={cleanLink} />);
 
     expect(component).toMatchSnapshot();
   });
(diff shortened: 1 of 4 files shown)


packages/node_modules/@webex/react-container-activity-list/src/formatters/at-mention.js
```diff
@@ -53,7 +53,7 @@ function AtMentionComponent({
       className={classNames(styles.atMention)}
       // eslint-disable-reason content is generated from elsewhere in the app
       // eslint-disable-next-line react/no-danger
-      dangerouslySetInnerHTML={{__html: content}}
+      dangerouslySetInnerHTML={{__html: content.replace(/<a(?![^>]*target="_blank"|\s*href\s*=\s*["'](?:mailto:|tel:))([^>]*)>/gi, '<a target="_blank"$1>')}}
```
bhabalan, line 56: This component is used when there is an @mention present in the message

packages/node_modules/@webex/react-component-activity-text/src/index.test.js
```diff
@@ -20,6 +21,12 @@ describe('ActivityText component', () => {
     expect(component).toMatchSnapshot();
   });
 
+  it('renders clean links properly', () => {
+    const component = renderer.create(<ActivityText content={cleanLink} />);
+
+    expect(component).toMatchSnapshot();
+  });
```
mkesavan13, line 28: Can we also have tests for cases where there is already `_blank` or `mailto:` or `tel:` in the anchor tag?

packages/node_modules/@webex/react-container-activity-list/src/formatters/at-mention.js
mkesavan13: Where does @mentioned links take to today in the widgets? Should we have a test case for that one as well?

mkesavan13 (changes requested):

packages/node_modules/@webex/react-container-activity-list/src/formatters/at-mention.js
bhabalan: The @mentioned links aren't clickable. They add a new tag called <spark-mention /> and wrap the person data around it. Whenever a mention is present in a message, this component gets used instead of react-component-activity-text. So I had to add the replacement logic here as well.
eg:
`<spark-mention data-object-type="person" data-object-id="801bb984-343b-4f6b-97ae-d13c91d4b877">John Doe</spark-mention>`

packages/node_modules/@webex/react-component-activity-text/src/index.test.js
bhabalan: Added tests

Shreyas281299 (approved):

mkesavan13 (approved):
Not sure if tests are there already and are required for the `@mentions` case. Please add if required. Approving

PR #1442, opened 2024-10-22
tooling(github-actions): add-pull-request-yml

Shreyas281299:
Completes Adhoc

This PR adds github action's checks in react-widgets. GIthub actions is a necessary now because circleCI builds are getting longer and they are failling because we dont have enough resources.

Working github actions run - https://github.com/Shreyas281299/react-widgets/actions/runs/11453800566

These github actions checks are specifically written for yarn workspaces.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,216 @@
+name: Pull Request CI
+run-name: ${{ github.actor }} is running Pull Request CI
+
+on:
+  pull_request_target:
+    types: [opened, labeled, reopened, synchronize]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  initialize:
+    runs-on: ubuntu-latest
+    env:
+      AWS_BUCKET: code.s4d.io
+      AWS_REGION: us-east-1
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20.13.1
+
+      - name: Clean yarn cache
+        run: yarn cache clean
+
+      - name: Initialize dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,216 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20.13.1
+
+      - name: Uncache Dependencies
```
adhmenon, line 53: Small doubt regarding the label name here, are we not caching the node modules here?

adhmenon (approved):
Overall changes LGTM. Just one comment on the `uncache` label.

.github/workflows/pull-request.yml
Shreyas281299: Caching is happening at line 34.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,216 @@
...
+
+      - name: Initialize dependencies
+        run: yarn install
+
+      - name: Cache Dependencies
+        uses: actions/cache@v3
+        if: steps.validate-dependencies.outputs.cache-hit != 'true'
+        with:
+          path: "**/node_modules"
+          key: node-modules-${{ hashFiles('./yarn.lock') }}
```
mkesavan13, line 39: Are we on yarn already during this PR?

mkesavan13 (changes requested):
Changes seem good except for the yarn-related ones. When this merges, we do not have yarn there already right? Is this in an assumption that we will have yarn changes in the subsequent PRs?

mkesavan13 (approved):

.github/workflows/pull-request.yml
Shreyas281299: We are not yet on yarn. Since this will go into a feature branch, and the very next PR that will go in is to convert the repo into a yarn workspace. Thats why to avoid doing the same task again, I have assumed that we are on yarn already

